PR #1613, opened 2013-04-08
Override Kunena templates in Joomla templates

obsidev:
The patch check if there is an override in the current Joomla template for the calling Kunena view.


level420:
Hi all!

does this still work for Kunena 4.0.3?

Thank you for your help.

Regards
Dietrich

810:
yes, this still works, but remember he is taking the old blue eagle code. Not the new hmvc code

level420:
@810: Thank you for your answer. 

Then: is there a similar or general way of overriding kunena templates in the site template?

Our task is simple: just write a h2-tag at the top of every kunena view, thus simulating the page title which comes from the the main menu entry.

But we struggle with this task and don't want to clone the complete kunena template for this simple thing.

PR #2453, opened 2014-05-06
K3.1 Add more options gender

810:


GJSchaller:
Suggestion - also add a choice to disable the Gender field completely, for communities that don't want to display this.

(I did some research on this for my own sites, the list is quite long, and varies from site to site.  I'm also not sure how to handle icons for all the choices...)

GJSchaller:
Other thought - allow the end user to enter it.

PR #2629, opened 2014-09-05
K3.1 [NFR]: Teaser #2354

810:
hides message for guests


coder4life:
@810 Does this support configuring how many replies that can be seen per topic.  What if we wanted to show a few replies after?  Also is there a login button to direct people, maybe where the topic should be.

810:
If enabled, you see only the first message. all replies are hidden.
for login button, we have now: https://github.com/Kunena/Kunena-Forum/pull/2710

components/com_kunena/site/template/crypsis/layouts/message/item/default.php
```diff
@@ -32,7 +32,13 @@
 <div class="badger-left badger-info <?php if ($message->getAuthor()->isModerator()) : ?> badger-moderator <?php endif;?>"
 	 data-badger="<?php echo (!$isReply) ? $this->escape($avatarname) . ' created the topic: ' : $this->escape($avatarname) . ' replied the topic: '; ?><?php echo $message->displayField('subject'); ?>">
 	<div class="kmessage">
-		<div class="kmsg"> <?php echo $message->displayField('message'); ?> </div>
+		<p class="kmsg">
+			<?php  if (!$this->me->userid && !$isReply) :
```
mahagr: These if statements are more complicated than they need to be.

    if (this->config->teaser && $isReply && !$this->me->userid) { teaser } else { message }

Another question is if we should hide the whole message and not just contents...

mahagr:
I'm not sure if the feature is complete and fully works yet as you can see messages from other contexes as well (search and so on).

810:
i have added the function on topics and search results. The only way to see the messages is to use the rss. There you still see the messages, i think.

mahagr:
Yes, and I think you can see them in a few other places as well. We'll see.

PR #2659, opened 2014-11-13
Fixed pagination issue in category view.

kavadas:
Pagination is broken when you are browsing a list of topics. 

How to reproduce the issue:
1. Go to a category with more than one page of topics. 
2. Use the pagination to jump to another page, let's say page 3.
3. Use the pagination again to go back to the first page and you will see again the page 3.

This is caused because the category model is setting the limistart value using the getUserStateFromRequest function. There is no need to fetch this value from session. This is something that is used usually in the administration lists where the pagination performs a POST request when you navigate through pages. In front-end the pagination is made using GET requests with the "start" parameter in the URL. When the user goes back to first page the start parameter is ( correctly ) not provided by Joomla! pagination and the getUserStateFromRequest returns ( correctly ) the last value from session.

### DIFF
--- a/components/com_kunena/site/models/topics.php
+++ b/components/com_kunena/site/models/topics.php
@@ -111,7 +111,7 @@ class KunenaModelTopics extends KunenaModel {
 		//$value = $this->getUserStateFromRequest ( "com_kunena.topics_{$active}_{$layout}_{$mode}_list_ordering", 'filter_order', 'time', 'cmd' );
 		//$this->setState ( 'list.ordering', $value );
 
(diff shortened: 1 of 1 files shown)


Hoffi1:
Hi guys,

we at JEM-project had the same problem and also hoped it will be fixed soon on Joomla site.
But in the meantime I decided to add a workaround. You can take a look here: https://github.com/jemproject/JEM-Project/commit/f6a35e104a7bbb69eae6963df84317879175cb0a

Maybe it's an idea for Kunena too. :smile:

Issue #2645, opened 2014-10-10
[master][3.0.6] Ebay doesn't work anymore


810:
Look like they changed their api support. We can't see any item anymore.

PR #2673, opened 2014-12-04
 [master][3.0.6] Ebay doesn't work anymore #2645 #2654 

xillibit:


xillibit:
What @josh was you thinking by saying, if i can't AJAX request i don't know what i should do :

> @xillibit can we move the request to the actual page somehow. Server should not be trying to fetch this but the individual browser. If for security reason this is not possible we need to discuss this more. This is the same problem as the gravatar issue as we could have many ebay items that have to be requested by the server. Kunena pages that have ebay items and ebay is slow will cause our page to slow down.

coder4life:
Yeah, need to think about this more

PR #2700, opened 2014-12-31
[3.1] Crypsis: Improvements

810:
* Fixed some mobile views
* Removed some inline css
* General improvements
* Improve the Credits page
* Update topic view
* Add breadcrumbs back to top
* reformat code on general.less
* Improve Stats view
* new feature template: trim message subject in message top
* Update message views
* Add custom.less support


810:
Can be merged now

components/com_kunena/site/template/crypsis/layouts/category/item/default.php
```diff
@@ -38,106 +38,112 @@
...
+			<ul class="inline">
+				<?php if ($categoryActions) : ?>
+					<li class="hidden-phone">
+						<?php echo implode($categoryActions); ?>
+					</li>
+				<?php endif; ?>
+			</ul>
+		</div>
+		<table class="table table-bordered">
+			<?php if (!empty($this->topics)) : ?>
```
coder4life: I would like to keep the behavior of keeping the table headers for when no topics are displayed.  I need to add a custom button for a suggestion to add a new topic.

The reason I want to keep the table headers is for learn-ability reasons, so people know ahead of time what to expect.  Please keep the old behavior as that was done with an intention of design.

810: I have moved them back because, Public write isn't working anymore, without this, The message actions are gone.

components/com_kunena/site/template/crypsis/layouts/credits/default.php
```diff
@@ -15,7 +15,7 @@
 	<?php echo JText::_('COM_KUNENA').' - '.JText::_('COM_KUNENA_CREDITS_PAGE_TITLE'); ?>
 </h2>
 
-<div class="well well-small">
+<div class="well well-small" id="Credits">
```
coder4life: lowercase ids

components/com_kunena/site/template/crypsis/layouts/category/item/default.php
coder4life: Not sure I understand, the message actions are above, and should not affect anything.
This if is only for the render of the table headers "Subject", "Reply", and "Last Post".

components/com_kunena/site/template/crypsis/layouts/message/item/top/default.php
```diff
@@ -17,21 +17,25 @@
 $attachments = $message->getAttachments();
 $avatarname = $this->profile->getname();
 $topicStarter = $this->topic->first_post_userid == $this->message->userid;
+if (KunenaConfig::getInstance()->ordering_system == 'mesid') {
```
coder4life: good practice to leave long variable usage like this in a variable at the top.

in other pages in practice we are using a variable like $config = KunenaConfig::getInstance()

Then in the actual line we are using $config->orderingsystem

components/com_kunena/site/template/crypsis/less/general.less
```diff
@@ -766,3 +779,57 @@ div.kmsg > li {
...
+	.btn-link {
+		color: inherit;
+	}
+}
+
+small.text-muted {
+	margin-top: -10px;
+}
+
+#Credits .dl-horizontal dt {
```
coder4life: same thing as above, lower case ids please

components/com_kunena/site/template/crypsis/layouts/category/item/default.php
810: New topic is missing now, when not login and empty category.

coder4life: yeah, this is why we need the filter work as it helps this problem

810: i have also fixed some other issue's on this file, i i revert it then all is gone, and i don't know what i fixed there. The best is to merge this pr, and then change it to your filters.

coder4life: well I would just remove this if condition and fix it to where it was before, do not revert entire file

Issue #2121, opened 2013-11-17
[develop][3.1.0]: Let user to choose if he wants display results by topics in advanced search  


xillibit:
Instead to show all messages in search results, let user choose if he want only see in results the topics

810:
this is merged, im closing this issue

PR #2709, opened 2015-01-01
[K3.1] Search: increase default date time option

810:
https://github.com/Kunena/Kunena-Forum/issues/2121


coder4life:
We cannot actually use this commit for security and performance reasons.  If you notice even Google does not allow queries over a certain time period past 365 days as seen here:
https://www.google.com/?gws_rd=ssl#q=google&tbs=qdr:y
What google does allow is a custom range over a certain time period, even older then 365 days.  We should be able to set a custom range with that range having a maximum of 1 year.

810:
I'm disagree with you. This is not a security or performance issue. I set it to 365 days as default, so users can choose other time dates, if they want. Now they must edit the file on each release. 

If you have a big website, you should stay on 365. 

@mahagr are you agree with me?

coder4life:
@mahagr is actually the one that advised me on this.  I agree it could be an option, however before I accept this commit please make a descriptive warning about setting dates too high, as the reason I closed this pull request is for the fact that people could shoot themselves in the foot.  This is one of those options where we want to discourage potential performance bottlenecks people could create for themselves.

810:
i have updated the pr. can i merge it?

PR #2712, opened 2015-01-02
[K3.0] Fix broken topic recounting #2586 #2627

810:
Fix for message recount on moderation


810:
I have fixed the recount issue, it was missing brackets and wrong position of if function.

@rich20 could you test it.

rich20:
Unapproved posts goes automatically in the trash and you can't restore.
"You do not have permissions to access this page."
The second problem is the creation date. All unapproved topics are
"Topic started 45 years 1 day ago by"

Issue #2711, opened 2015-01-02
[K3.1] Ban: logout user when banning the user


810:
When banning a user, the user should be logout . For cleaner who is online stats.
Now the user stays logged in, until the session is expired, can be days until the user is gone.

PR #2715, opened 2015-01-02
[K3.1] Ban: logout user when banning the user #2711

xillibit:


810:
Could you change it to only the site option ban.

PR #2717, opened 2015-01-02
Youtube video doesn't work because links aren't correct

xillibit:


810:
only autoembedyoutube is working the video's from video tags are broken "This content cannot be displayed in a frame". And the video of the issue, you get "An error is occurred, please try again later"

810:
You can close this issue, because the user had a extra character in his url, so everything still works.

xillibit:
If you put link into video brackets like that : [video]https://www.youtube.com/watch?v=bpk7ERlZ0CI[/video]  the video isn't displayed, but yes if put only link it works

810:
if you have only "[video width=425 height=344 type=youtube]bpk7ERlZ0CI[/video]" you get Cannot save message: Message body is empty!

xillibit:
That's was an another issue, i don't find it anymore in issues list

810:
@test +1

rich20:
> if you have only "[video width=425 height=344 type=youtube]bpk7ERlZ0CI[/video]" you get Cannot save message: Message body is empty!

It's the same with this bbCode :  [video]https://www.youtube.com/watch?v=bpk7ERlZ0CI[/video] 
Without text you cannot save videos. Enter a text and both code work.

810:
yeah that is a different issue, is the YouTube still working?

rich20:
> yeah that is a different issue, is the YouTube still working?

- [x]  YouTube works

Issue #793, opened 2012-06-04
Cannot hide admins in userlist


sozzled:
See http://www.kunena.org/forum/168-K-17-Support/123164-Remove-user-from-Userlist#125445 for details

xillibit:
I have re-tested it now, it's works well with K3.0.x but with K3.1.0 it doesn't work i need to check what's going on

rich20:
It works also with K.1.0 but vice versa in Blue Eagle.
In Blue Eagle you must set "Show Super Users in user list = Yes", then he is hidden. 
In Crypsis it works correct.

xillibit:
Can-you write here : https://github.com/Kunena/Kunena-Forum/pull/2718 if it's working now ?

810:
Is now fixed, is now in K3.1, Im closing this issue

Issue #2426, opened 2014-04-02
Option "Show E-mail" appears in both Users and Extra tab in Configuration


GJSchaller:
The option "Show E-mail" appears in both the Users tab, and the Extras tab, in the Configuration section of Kunena Administration.  This is redundant and confusing, and does not indicate which one works (or if either / both do, which has priority, etc.).

810:
i'm closing this issue, the strings are correct, and in the right position.

PR #2718, opened 2015-01-03
[k3.1] Regression: setting to hide admins in userlist has no effect #793

xillibit:
Needed to remove the setting "Show Super Users in user list" under  tab Users because was duplicate (already available under tab Extra)


coder4life:
Do we know the commit that created the regression?

xillibit:
I had done a fix for it but only for blue eagle into K3.1 and not for crypsis so it's not really a regression

810:
or https://github.com/Kunena/Kunena-Forum/issues/2426

rich20:
In Blue Eagle it's vice verca. You must set "Show Super Users in user list = Yes", then he is hidden.
In Crypsis it works correct.

Issue #2442, opened 2014-05-04
[develop][3.1.0]: Load poll-min.js only on topic page


810:
We need to load the script only on the topic page. not on all pages. 
* Improve speed

810:
merged

PR #2729, opened 2015-01-14
[K3.1] Speed improvement: Move js to layout

810:
Move the javascript loading to the right views. So it doesn't loaded on
each page.

https://github.com/Kunena/Kunena-Forum/issues/2442


components/com_kunena/site/template/crypsis/template.php
```diff
@@ -82,63 +82,20 @@ public function loadLanguage()
 	 */
 	public function initialize()
 	{
-		// Template requires Mootools 1.4+ framework.
-		$this->loadMootools();
```
xillibit: Untill we have not replaced attachments uploader and bbcode editor, we still need to load mootools

components/com_kunena/site/template/crypsis/layouts/search/results/default.php
```diff
@@ -10,7 +10,35 @@
...
+{
+	$template = KunenaTemplate::getInstance();
+	if ( $template->params->get('lightboxColor') == 'white') {
+		$this->addStyleSheet('css/fancybox-white.css');
+	}
+	else  {
+		$this->addStyleSheet('css/fancybox-black.css');
+	}
+	$this->addScript('js/fancybox.js');
+	JFactory::getDocument()->addScriptDeclaration('
```
xillibit: The problem by doing that is that the same code is present in two differents places, it's difficult to maintain

components/com_kunena/site/template/crypsis/template.php
810: Mootools is still loaded. So its a unnecessary load request.

PR #2731, opened 2015-01-16
PNG Image Files Optimization

Dade88:
Lossless optimization of ALL possible Kunena PNG files to a smaller
version.

~ 80 KB Saved.


810:
thank you, i tried yesterday the same, only for me some images where bigger then we all already got. 

i tried http://www.punypng.com/

Dade88:
PNGOut is my fav (command-line tool), but you can use many others tools (PNGGauntlet uses multiple tools and serves you the best option).

Also Kraken.io turns useful (I found it's better than smush.it) and of course, "Save for Web" plugin for GIMP is a must-have.

Issue #2704, opened 2014-12-31
[K3.1] Crypsis: Hide announcement on new topic page


810:
hide the announcement on the new topic page, to improve the user experience.

810:
merged

PR #2735, opened 2015-01-19
[K3.1] Crypsis: Hide announcement on new topic page #2704

810:
also excluded on profile and on announcement


components/com_kunena/site/controller/widget/announcement/display.php
```diff
@@ -45,6 +45,13 @@ protected function before()
 			return false;
 		}
 
+		$view = JRequest::getWord ( 'view', 'default' );
+		$layout = JRequest::getWord ( 'layout', 'default' );
```
xillibit: JInput should used here instead of use JRequest, because JRequest will be replaced by JInput call

Issue #1476, opened 2013-03-12
No button for [HR] in editor


GJSchaller:
There is an {hr} (Horizontal Rule) function in the BB Code, but no button for it in the current editor.

810:
merged

mahagr:
Unfortunately the change removed transparency, reopening issue.

810:
fixed and merged

PR #2736, opened 2015-01-19
[K3.1] Crypsis: No button for [HR] in editor #1476

810:
also added tweet icon


mahagr:
@810 You have removed transparent background from the images!!!

PR #2744, opened 2015-01-20
[3.1] Crypsis : Show E-mail = No has no effect. The address is always public #2742 

810:
#2742


rich20:
Show E-mail = No : 
- [x] Administrators can see it
- [x] Members cannot see it
- [x] Guests cannot see it

Show E-mail = Yes - User setting in the profile : Hide E-mail = Yes
- [x] Administrators can see it
- [x] Members cannot see it
- [x] Guests cannot see it


Show E-mail = Yes - User setting in the profile : Hide E-mail = No
- [x] Administrators can see it
- [x] Members can see it
- [x] Guests can see it

Issue #2545, opened 2014-07-04
[develop][3.1.0] "Cleaning" button for spammed users in backend


Mortti:
Delete all "Lifetime" banned users at ones from Kunena and Joomla! database...

810:
merged

PR #2746, opened 2015-01-20
[3.1.0] "Cleaning" button for spammed users in backend #2545

810:


xillibit:
The ban part could change (if PR is merged) with that PR : https://github.com/Kunena/Kunena-Forum/pull/2390

Issue #1550, opened 2013-03-30
[master][2.0.X]: When session is expired on backend you get some errors


810:
When you on edit template, you get some errors:

Notice: Trying to get property of non-object in components/com_kunena/template/joomla25/templates/edit.php on line 35
Notice: Trying to get property of non-object in components/com_kunena/template/joomla25/templates/edit.php on line 39
Notice: Trying to get property of non-object in components/com_kunena/template/joomla25/templates/edit.php on line 47

1) go to backend-kunena templates- edit template
2) session expired
3) login again
4) you redirected to the same page (edit templates) but with errors

PR #2752, opened 2015-01-22
[K3.0]: When session is expired on backend you get some errors #1550

810:


rich20:
The errors are gone now. But after login again, the template_thumbnail.png  will not be showing. But I think, that's a insignificant problem.

810:
yeah, I thought that too

PR #2754, opened 2015-01-23
[K3.0] Language updates

810:


mahagr:
Can we revert this change and redo it without breaking en-GB and adding two non-working languages: cs_CZ and eu?

There's also some issues with Korean language.. A lot of translations get reverted to English in it.

PR #2766, opened 2015-01-28
[3.1] Rss: Hide message if teaser is enabled

810:
- by posts
- by recent messages

### DIFF
--- a/components/com_kunena/site/views/topics/view.feed.php
+++ b/components/com_kunena/site/views/topics/view.feed.php
@@ -167,7 +167,12 @@ class KunenaViewTopics extends KunenaView {
 			} else {
 				$id = $topic->last_post_id;
 				$page = 'last';
-				$description = $topic->last_post_message;
+				if (!$this->me->userid && $this->config->teaser && $id != $topic->first_post_id) {
+					$description = JText::_('COM_KUNENA_TEASER_TEXT');
+				}
+				else {
+					$description = $topic->last_post_message;
+				}
 				$date = new JDate($topic->last_post_time);
 				$userid = $topic->last_post_userid;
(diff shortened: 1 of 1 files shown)


rich20:
- [x] By Posts - Success
- [x] By Recent Messages - Success

xillibit:
Works only with blue eagle with two RSS settings, with crypsis it works only for first message of RSS feed

sozzled:
When you think about it, if a site owner enables RSS feeds but effectively says, "here's the RSS feed but, no, I'm not showing anyone who is not logged in what content has been added", this actually contradicts the whole point of allowing RSS feeds, doesn't it?

Sure, it might be possible to restrict the content in RSS messages but, really, when you think about it, if a site owner "teased" everything then the RSS would have nothing to report.  The simplest way to overcome the "problem" is to disable RSS feeds.

I think that allowing RSS feeds on a site that has K 3.1 where the teaser feature is enabled will result in more questions (and possible complaints) being raised that it will benefit users by "fixing" the source code in the way that was suggested by Jelle.

Just my opinion.

rich20:
The rss-feed makes sense if you use "Feed type By Topic". It shows only the first message from a topic (the same as with teaser in the forum).

sozzled:
It's just my opinion, of course, but I still feel that you have not thought through the ramifications of allowing RSS feeds but, effectlvely, disabling the content.  Remembering that the purpose of teasers is to "encourage" people to login to the forum to use it, RSS feeds (because there is no "login") is a back-door way of accessing the forum.

The problem could be addressed with a techical solution (as proposed by Jelle) but, in practice, having RSS feeds and teasers is somewhat contractory and defeats the purposes of both.

PR #2774, opened 2015-02-14
Lightbox can't be open on textlink of attachments when protected is on

xillibit:


810:
didn't work for me

xillibit:
In which case it doesn't work ?

mahagr:
I'm not fine with this change. Think of photography forum with a lot of thumbnail images in it. It also doesn't work with protected messages if SEF is turned off.

We need to fix lightbox script instead and remove file type check from it.

PR #2777, opened 2015-02-14
[K3.1] Subscription : The option "Check Subscription Box = Yes or No" sh...

810:
...ould be in the profile available

https://github.com/Kunena/Kunena-Forum/issues/2759

Issue #2489, opened 2014-05-13
[develop][3.1.0]: Let the user set the item selected by default on recent discussions and search views


xillibit:
On recent discussions page on the time filter and on search page add options to let user set the item selected by default

810:
merged

PR #2778, opened 2015-02-15
[K3.1]: Let the user set the item selected by default on recent discussions and search views #2489

810:
merge this one after the #2777


components/com_kunena/site/controllers/user.php
```diff
@@ -573,6 +573,7 @@ protected function saveSettings() {
 		$this->me->hideEmail = JRequest::getInt('hidemail', '');
 		$this->me->showOnline = JRequest::getInt('showonline', '');
 		$this->me->canSubscribe = JRequest::getInt('cansubscribe', '');
+		$this->me->userListtime = JRequest::getInt('userlisttime', '');
```
xillibit: You should use JInput isntead JRequest when modifing/adding some parts of code

rich20:
### Backend
- [x] Selected time period for Search - Success

### Frontend
- [x] Profile:  Select time period for Recent Topics - Success
- [ ] Recent Topics: Temporary Time Period - Failed (The time periods are permanently defined. You can't change the time period temporarily at the Recent Topics page.)

810:
could you check if the model has changed for you:
components/com_kunena/site/models/topics.php 

Also check if the setting is saved

rich20:
Maybe you misunderstood me. The profile setting works. But there is also a drop down menu at the Recent Topics page. And here you can't switch.

810:
ok, i thought you getting an error. But the switch doesn't work anymore, is always the user setting. @xillibit do you have any idea, to override this

810:
@rich20 i think i fix it, could you check it

810:
closing this one: see https://github.com/Kunena/Kunena-Forum/pull/2800

rich20:
It's not optimal. If you switch on recent topic page it is only temporary. After page refresh it switch back to the user setting. The same if you go to the next page.

PR #2782, opened 2015-02-17
[K3.0.8] Fix jomsocial activity errors

810:


xillibit:
What errors do-you have ?

810:
http://www.kunena.org/forum/k-3-0-fatal-errors/132425-500-internal-server-error-started-with-no-changes-to-website

xillibit:
It's not fatal error, it's just strict standard errors

810:
if you fix the standards, you get a fatal. on line 94

xillibit:
The KunenaForumMessage cast that you remove has been added previously by matias (don't know exactly the reason) : https://github.com/Kunena/Kunena-Forum/commit/5b6a63b2ac5841c7e3f37bd92d46b4fd3befa27d

components/com_kunena/admin/install/plugins/plg_kunena_community/activity.php
```diff
@@ -91,7 +91,7 @@ public function onAfterReply(KunenaForumMessage $message) {
 		$act->cmd = 'wall.write';
 		$act->actor = $message->userid;
 		$act->target = 0; // no target
-		$act->title = JText::sprintf('PLG_KUNENA_COMMUNITY_ACTIVITY_REPLY_WALL',$params->get('actor_url'),$params->get('actor'),$params->get('title_url'),$params->get('title'));
```
mahagr: I believe that this change came from JomSocial team updating our code to their new standards.

810: it was missing language string. and the code doesn't work. 

So i reverted it back. and the code is still working on the lasted jomsocial version

PR #2784, opened 2015-02-18
Crypsis : zip icon isn't displayed on attachments thumbnail

xillibit:


810:
+1

mahagr:
This change breaks all file uploads.

PR #2728, opened 2015-01-11
 Crypsis : implement blue imp jquery uploader to handle attachments #2641

xillibit:

PR #2790, opened 2015-02-20
jQuery uploader: when protected is on on edit attach image isn't showed

xillibit:


810:
+1

rich20:
- [x] Success

It seems also the problem with the lightbox is solved. https://github.com/Kunena/Kunena-Forum/pull/2728

mahagr:
Doesn't really fix the issue; need to be reverted and fixed in lightbox instead.

Issue #2140, opened 2013-11-23
Crypsis : You can create only new topics, if a topic already exists


rich20:
You can't create new topics in empty categories. If the category is empty, the button New Topic is missing. Then it works only via the tab "New Topic".

PR #2793, opened 2015-02-20
Crypsis: You can create only new topics, if a topic already exists #2140

xillibit:


rich20:
- [x] Success

810:
+1

mahagr:
Changes that change every single line should have never been accepted. We really need to fix that CRLF issue in our files after merge...

mahagr:
Checked the change locally and it's ok...

Issue #2794, opened 2015-02-20
[develop][3.1.0] The profile option "Select subscription as default" has no effect in quick reply


rich20:

810:
should be fixed now, closing this issue

PR #2799, opened 2015-02-22
[K3.1] Crypsis:  The profile option "Select subscription as default" has no effect in quick reply #2794

810:


810:
@rich20 please try now.

810:
try again now

PR #2800, opened 2015-02-22
[K3.1]: Let the user set the item selected by default on recent discussions and search views #2489

810:
...

part2


xillibit:
If i set the limit in the profile to 4 hours in recent discussions page it show 4 hours too, it doesn't take care of what you have set in Kunena config

810:
That's the point of this Pr, So users can handle it by there self

xillibit:
What's the purpose to have a setting in Kunena configuration for that ?

810:
For guests, and the normal users.
Some users want to change some settings, this is 1 of them.

I think 96% will use the standard kunena configuration.

PR #2801, opened 2015-02-22
Fix #1367 topic recounting

### LINKED ISSUE #1367: Category with Review posts
I just notice that when you do category what need review, then new Topics are like a Replies?
See: http://screencast.com/t/E1e4ZrrSd

When you log in like SU or Moderator they are still.
See: http://screencast.com/t/WxZdSbx4OJOV

2 Topics... see:
http://screencast.com/t/R69uCED7TP26

After you approve success... those are Topics(not Replies)!!!
See: http://screencast.com/t/15AMm5vh

Tested:
J! 2.5.8
Kunena 2.0.4 Stable

mahagr:
Should fix topic recounting on messages that are under review.


mahagr:
There are still issues with this fix, but it seems to do something right..

Issue #2817, opened 2015-03-09
[develop] Crypsis : You can only insert one map per page


rich20:
Each additional map shows "Unable to display Google Map.". But the city will in addition shown on the first map.

rich20:
- [x] Success

PR #2818, opened 2015-03-09
[develop] Crypsis : You can only insert one map per page #2817

xillibit:


rich20:
- [x] Success

810:
- [ ] Failed: Unable to display Google Map.

If you insert 2 maps in 1 message. Both locations are loaded in the first message

rich20:
Sorry, you're right. I've tested only with one map per post.

components/com_kunena/site/template/crypsis/layouts/bbcode/map/default.php
```diff
@@ -24,7 +24,7 @@
 	$id = 0;
```
810: Please update this to (New look/api, code still work): 
$this->addScript('http://maps.google.com/maps/api/js?v=3.exp&signed_in=true');
$id = 0;

810:
When i do
```` ->set('mapid', ++ $bbcode->parent->id)````
The maps are loaded correct, but the action buttons are gone also moderate is gone

**Comment by xillibit** [comment]
You can try now, i have updated it

**Comment by 810** [comment]
Code work, but please fix codestyle for : var $mapid=0; and $this->mapid++;

**Comment by xillibit** [comment]
Done

File: libraries/kunena/bbcode/bbcode.php [hunk 2]
```diff
@@ -328,6 +328,9 @@ public function getEbayItemFromCache($ItemID)
  */
 class KunenaBbcodeLibrary extends BBCodeLibrary {
 	var $default_smileys = array();
+
+	public $mapid=0;
```
810: ```` public $mapid = 0; ````

xillibit:
Corrected

810:
:+1:

Issue #2837, opened 2015-03-16
[develop][3.1.0] User Edits : We should add a third option


rich20:
We have now:
User Edits = "Yes" (with time setting)
User Edits = "No"
But it's not always optimal, when users after answer edit his previous post.

The third option should be:
User Edits = "Yes, except no answers"

810:
its merged

PR #2838, opened 2015-03-16
[K3.1] User Edits : We should add a third option #2837 

810:
https://github.com/Kunena/Kunena-Forum/issues/2837


xillibit:
Why we should add a third option ?

Issue #2803, opened 2015-02-22
Crypsis: All file downloads are broken


mahagr:
Pull #2784 breaks all downloads.

810:
fixed and merged

PR #2848, opened 2015-03-19
[K3.1] Crypsis: All file downloads are broken #2803

810:


xillibit:
Need to be tested with protected attchments disabled and enabled, and too with SEF enabled and disabled

xillibit:
It works for me :+1:

Issue #2748, opened 2015-01-21
[K3.1] Crypsis: index Notice: Uninitialized string offset: 1


810:
Notice: Uninitialized string offset: 1 in C:\wamp\www\master\libraries\joomla\filesystem\path.php on line 209

Call Stack

#

Time

Memory

Function

Location

1 0.0005 242896 {main}( ) ..\index.php:0 
2 0.0642 3500968 JApplicationCms->execute( ) ..\index.php:40 
3 0.0642 3501168 JApplicationSite->doExecute( ) ..\cms.php:251 
4 0.1472 8393368 JApplicationSite->dispatch( ) ..\site.php:230 
5 0.1534 8486608 JComponentHelper::renderComponent( ) ..\site.php:191 
6 0.1555 8544880 JComponentHelper::executeComponent( ) ..\helper.php:360 
7 0.1560 8613024 require_once( 'C:\wamp\www\master\components\com_kunena\kunena.php' ) ..\helper.php:380 
8 0.6316 10676408 KunenaControllerApplicationDisplay->execute( ) ..\kunena.php:83 
9 0.7176 13117216 KunenaLayout->render( ) ..\display.php:105 
10 0.7177 13117424 KunenaLayoutBase->render( ) ..\layout.php:60 
11 0.7185 13139192 include( 'C:\wamp\www\master\components\com_kunena\template\crypsis\pages\category\list\default.php' ) ..\base.php:165 
12 0.7185 13139320 KunenaLayoutPage->execute( ) ..\default.php:13 
13 0.7198 13206344 KunenaControllerDisplay->execute( ) ..\page.php:46 
14 0.7198 13207720 ComponentKunenaControllerCategoryIndexDisplay->before( ) ..\display.php:29 
15 0.7284 13997448 KunenaForumCategory->getRSSUrl( ) ..\display.php:140 
16 0.7284 13997704 KunenaRoute::_( ) ..\category.php:248 
17 0.7284 13997880 KunenaRoute::prepare( ) ..\route.php:108 
18 0.7298 14199880 KunenaRouteLegacy::convert( ) ..\route.php:475 
19 0.7300 14202120 JLog::add( ) ..\legacy.php:548 
20 0.7301 14203648 JLog->addLogEntry( ) ..\log.php:151 
21 0.7322 14267704 JLogLoggerFormattedtext->addEntry( ) ..\log.php:262 
[...]

PR #2865, opened 2015-03-24
[K3.1] Crypsis: index Notice: Uninitialized string offset: 1 #2748

810:


xillibit:
It's in Joomla! code that check on folder need to be placed because it's in JLog class that it's handled

Issue #1530, opened 2013-03-27
Add Column / Filter for "Rank Title" in User Manager


GJSchaller:
When viewing the list of users, it would be extremely useful to be able to filter and sort on Rank Title (Administrator, Banned, Spammer, etc.) in the User Manager in the back end.

810:
merged

PR #2876, opened 2015-03-25
Add Column / Filter for "Rank Title" in User Manager #1530

810:


810:
@florian could you check the filter. because in the database is always 0

PR #2877, opened 2015-03-25
Remove jquery.ui from Kunena package and load it from Joomla!

xillibit:


810:
Do we need it at all?

Issue #41, opened 2011-07-09
Add category icon upload and gallery


fxstein:
Use uploader like in front end (post) to allow to upload icons/images for categories and perform resizing if the file exceeds the specified category icon size.

Do the same for an icon gallery: Let admins upload individual category images/icons.

Issue #2726, opened 2015-01-09
[3.1] Category Image Path: does not work


GJSchaller:
When "Config => Frontend => Category Image Path" is set in the Kunena config, and the icons are present...

Crypsis still pulls from:
/components/com_kunena/template/crypsis/media/topic_icons/default/system/(icon).png

Blue Eagle pulls from:
/media/kunena/topic_icons/default/system/(icon).png

xillibit:
Maybe this setting should be put directly for each template settings

GJSchaller:
I would agree with @xillibit

xillibit:
We should keep this setting in Kunena configuration settings because old templates may still use it. In templates shipped this setting shouldn't be used

GJSchaller:
This looks similar to #2262 - can we close old ones to get them out of the system?

810:
fixed and merged

PR #2881, opened 2015-03-26
Add category icon upload and gallery #41

810:
Instead of uploader, we should use input because now you can use icomoon
icons, or own classes

### DIFF
--- a/components/com_kunena/admin/install/install.xml
+++ b/components/com_kunena/admin/install/install.xml
@@ -136,6 +136,9 @@
 		<field name="alias"
 		       type="varchar(255)"
 		       null="0"/>
+		<field name="icon"
+		       type="varchar(20)"
+		       null="0"/>
 		<field name="icon_id"
 		       type="tinyint(4)"
 		       null="0"
--- a/components/com_kunena/admin/install/kunena.install.upgrade.xml
+++ b/components/com_kunena/admin/install/kunena.install.upgrade.xml
@@ -641,7 +641,7 @@
 				AFTER `userid`
 			</query>
 		</version>
-		<version version="4.0.0-ALPHA"
+		<version version="4.0.0-DEV"
 		         versiondate="2015-03-26"
 		         versionname="Internal">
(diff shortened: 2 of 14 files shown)


components/com_kunena/admin/install/install.xml
```diff
@@ -136,6 +136,9 @@
 		<field name="alias"
 		       type="varchar(255)"
 		       null="0"/>
+		<field name="icon"
```
xillibit: Instead of adding a new field in categories table, the setting of icon  should be stored into params field https://github.com/Kunena/Kunena-Forum/blob/develop/components/com_kunena/admin/install/sql/install/mysql/create_tables.utf8.sql#L74

810: how can i do that, whats the code and where

xillibit: I will take a look on the evening, matias in the past had said to user params row in category table instead to add a new row

coder4life: Actually it is the icon_id

GJSchaller:
How does this tie into #2726 - can they be the same issue?

components/com_kunena/admin/install/kunena.install.upgrade.xml
```diff
@@ -641,6 +641,13 @@
 				AFTER `userid`
 			</query>
 		</version>
+		<version version="4.0.0-DEV"
+		         versiondate="2015-03-26"
```
xillibit: You need to set for vesion : 4.0.0-ALPHA2 else the row icon isn't added in table

810:
updated the code

rich20:
The classes works now. But how to find users these icon gallery?

810:
they must find it by there template builder, for protostar: http://help.joomla.fr/3/index.php/tutoriels/bootstrap/165-bootstrap-icones-de-la-police-icomoon-de-joomla-jui-j-3-2-0

Issue #2816, opened 2015-03-09
[master][3.0.8] Menu item Recent Posts shows only own posts


rich20:
Menu item Recent Posts shows only own posts but guests see all posts. If you write something in the field "User", also members can see all posts.
![bildschirmfoto vom 2015-03-09 20 18 56](https://cloud.githubusercontent.com/assets/888847/6563364/ddb29d48-c6a0-11e4-85f5-e4c1f318eddf.png)

xillibit:
Is fixed this one ?

rich20:
Now se all users the recent posts.

rich20:
For me it's correct.

PR #2882, opened 2015-03-26
[K3.0.8] Menu item Recent Posts shows only own posts #2816

810:
Its must be userid, not username.

When login, own messages are showed, this is normal behavior i think.


rich20:
>Its must be userid, not username.

>When login, own messages are showed, this is normal behavior i think.


I mean, it's not optimal. You know the own posts. Users want to see also the other posts.

810:
that's the recent topics part

rich20:
hmm.. I don't understand this menu item.
Description: Recent Posts - Displays recent posts with a few options to choose from

810:
we need to fix it, in the mean time, i merge this language strings

Issue #1127, opened 2012-10-06
[master][2.0.x]: Add button for moderators: View poll results


810:
Add button for moderators: View poll results

PR #2893, opened 2015-03-28
[Crypsis] Add button for all users: View poll results #1127

xillibit:


rich20:
- [ ] If you visit this topic, the result is already open
- [ ] The button is visible for all users
- [ ] All users see the result

xillibit:
This change only when you see poll options in order to vote, if you have already voted to the poll you can see only the results.

It's intented that is available for all users : https://github.com/Kunena/Kunena-Forum/issues/1127#issuecomment-9268983, the name of pull request need to be changed...

rich20:
The result is open as default. Also if they have not voted. With the button you can close it.

xillibit:
@rich20 : try now i have updated it

rich20:
- [x] Success

PR #2903, opened 2015-04-01
Issue when putting wrong string for filter_order

xillibit:


810:
userlist is now empty

components/com_kunena/site/models/user.php
```diff
@@ -151,8 +151,16 @@ public function getItems() {
 			$query .= " ORDER BY {$db->quoteName($this->getState ( 'list.ordering'))} {$this->getState ( 'list.direction')}";
 
 			$db->setQuery ( $query, $limitstart, $limit );
-			$items = $db->loadColumn();
-			KunenaError::checkDatabaseError();
+			try
```
coder4life: Generally its not a good practice to have a try and catch block here. The best place to have try and catch blocks is on the state of the object. When an error gets thrown from the db because of loadColumn the object is in a bad state because the data was never retrieved.  I guess an analogy is your engine to your care is overheating, the engine will keep running until it seizes.  You will be given alerts that your engine is on fire, but the actual error about your engine block failure is not until the car stops working when it is expected to keep working.  This is not a case for every case, however we have a very simplistic call here which really does not need a try catch block for an error on input that other code should be checking.

xillibit:
It works for me : 

dsdfre

810:
When you add the code on the url, the list is empty until you click on order (username,posts,karma ect)

810:
you need turn sef on, and go to forum/user/list?filter_order=%22%27%3E%3Cqss%20%60%3b!--%3D%26%7b()%7d%3E&filter_order_Dir=asc&b5d08715e889dc5a1d8c5baae2e64201=1

810:
+1

Issue #2902, opened 2015-03-31
[develop] Crypsis : If you create a new topic the subject is already included.


rich20:
If you create a new topic the subject is already included. It's always the same title, as in the first topic.

rich20:
hmm.. it seems a cache problem. It's always the titel from the topic where was last written.

810:
Its the new feature, that saved data, for recovery and from quickreply to normal editor

rich20:
Now you get no warning if you forgot a topic title.

xillibit:
The last entry shouldn't appear in inputs if you submit your message

rich20:
You can easy test it. Read a message (no matter which one) and then create a new topic. In input field is the same subject like those of the last read message.

rich20:
I see now, the same problem gives with Reply. 
Example:
The topic is "Welcome to Kunena!"
Click on Reply and change the topic title for this one message to "Welcome to Kunena testpage"
now click at the first message on Reply
in the input field is "Welcome to Kunena testpage" and not "Welcome to Kunena!"

xillibit:
Is-it fixed now ?

rich20:
- [x] SCuccess

PR #2923, opened 2015-04-04
[develop]Crypsis: If you create a new topic the subject is already here #2902

xillibit:


810:
the data isn't saved from quickreply to editor

xillibit:
In which case it should be saved from quickrely to editor form ?

810:
untitled

xillibit:
I will ask to the author of script because i don't know why it doesn't work in this case

810:
it worked now, but can we remove the 10 sec or change it to 1 sec.

Also we need clear data on cancel or close qreply

xillibit:
I have updated the PR

> Also we need clear data on cancel or close qreply and on submit

On cancel or submit the data are already cleared automatically by the script

810:
Ok after testing, this works, should be
````
/* Store form data into localstorage for 2000 second */
	jQuery("#postform").sisyphus( {
			locationBased: false,
			timeout: 2000
	});
````

xillibit:
Why the timeout on 2000 seconds ? So each every 2000 seconds the data are stored in browser storage so it's a long time

810:
I think the option is How long store data. When i set it to 2000, the data is stored.

xillibit:
No :  ```timeout is an interval in seconds to perform saving data. If 0 - save on each time field is updated.```

xillibit:
PR updated

rich20:
Does not work. Go in your profile and empty the cache. Then try edit some message. The edit field will empty.

xillibit:
Branch updated

rich20:
- [x] Success

Issue #2854, opened 2015-03-21
[develop] Crypsis Profile : Time periods does not work (Recent Posts etc..) It goes back to All. 


rich20:

810:
@rich20 could you check the fix.

PR #2946, opened 2015-04-09
[K4.0] Crypsis: Time periods does not work (Recent Posts etc..) It goes back to All. #2854

810:
ps, change setting on kunenmenu, select item,  main settings tab, change
filter to global


810:
Only on profile, its stills "All"

PR #2958, opened 2015-04-12
Fix update list for 4.0 in install.php

xillibit:


install.php
```diff
@@ -225,7 +225,7 @@ protected function fixUpdateSite() {
 		$query = $db->getQuery(true)
 			->set($db->quoteName('name').'='.$db->quote('Kunena 3.1 Update Site'))
```
810: change this also

xillibit:
done

PR #2969, opened 2015-04-14
Don't show crypsis template onto Joomla! 2.5

xillibit:


rich20:
- [x] Success (Crypsis is no longer available)

PR #3001, opened 2015-04-18
[K3.0.9] Fix error message on none empty message

810:
before a normal user can't do

[confidential]some text[/confidential]

Then he get the error that the message is empty

### DIFF
--- a/components/com_kunena/site/controllers/topic.php
+++ b/components/com_kunena/site/controllers/topic.php
@@ -131,16 +131,6 @@ class KunenaControllerTopic extends KunenaController {
 			if ($file['error'] != UPLOAD_ERR_NO_FILE) $message->uploadAttachment($intkey, $key, $this->catid);
 		}
 
-		// Make sure that message has visible content (text, images or objects) to be shown.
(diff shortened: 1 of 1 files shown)


rich20:
In principle it works. But when a normal user edit his confidential informations, then the message will deleted.
Message
The message has been successfully deleted.

810:
Try now, 

I have removed the check on empty message, so users can't delete their topic. 
This fixed an other issue, that users can delete their message even if the setting is disabled.

But we need to add a function on backend at recount messages, with a new option: remove empty messages

rich20:
- [x] Success

> This fixed an other issue, that users can delete their message even if the setting is disabled.
But we need to add a function on backend at recount messages, with a new option: remove empty messages

I also think so. The admin should decide whether he wants to delete blank posts.

xillibit:
I'am not sure if it's good idea to do it in this way

xillibit:
You can't just post a message with just confidential because people with haven't the right to see the confidential will see an empty message

810:
closing this issue for now.

PR #3005, opened 2015-04-20
Issue on reply when you have in previous message a confidential tag

xillibit:
On reply if you have a previous message with a confidential item in it, you will see this error : 

```
Rendering Error in layout Topic/Edit/History: Property "userid" is not defined in C:\wamp\www\joomla3.4.0\libraries\kunena\bbcode\bbcode.php on line 1427
Layout was rendered in C:\wamp\www\joomla3.4.0\libraries\kunena\controller\display.php on line 161
```


810:
i don't have this issue

xillibit:
Create a new topic and in the message put a confidnetial tag, then reply to this topic and in the bottom instead of previous message you will see the error like in first message

810:
don't have this issue

rich20:
I don't have this error.

PR #3019, opened 2015-04-25
Remove last usage of deprecated CKunenaUpload class

xillibit:
This affect only attachments upload under blue eagle template


810:
Doesn't blue eagle need this?

xillibit:
I have replaced the code to don't use that deprecated class, blue eagle used this but not directly

810:
3rt party template still need this class. So we can't delete it yet

xillibit:
it was used in attachements class so it doesn't affect others templates

810:
ok, then I will merge this one

Issue #2387, opened 2014-03-17
  [master][3.0.7] Attachment icon is displayed without attachments


rich20:
If you (permanent)delete all messages with attachment in topics, it will still displayed the symbol. Also if you delete all attachments in backend it's the same. It works only, if you edit and delete the attachments directly in messages.
![attachment](https://f.cloud.github.com/assets/888847/2438325/5a37dc68-adf0-11e3-9c6b-0418b1971385.jpg)

810:
Seems fixed on K5.0, I have no longer this issue. @rich20 could you confirm?

rich20:
You must still manually count statistics.

810:
- [ ] Deleting topics from backend
- [ ] Deleting messages from backend
- [ ] Deleting permanently topics in frontend
- [ ] Deleting permanently messages in frontend
- [ ] Deleting attachments directly in messages
- [x] Delete attachment backend attachment manager
- [x] Delete attachment profile - attachment manager

810:
merged

PR #3068, opened 2015-05-09
When deleting attachment in backend update topic object #2387

xillibit:


rich20:
- [x] Success (it remains only the code in messages "Attachment not found")

The topic should also be updated, if you delete attachments in User-Profile Attachment Manager.

xillibit:
I haven't check the message i think i need to update it

xillibit:
You can try noow for backend and profile should be good

rich20:
- [x] Delete in backend - Success 
- [x] Delete in user profile - Success 

(It remains only the code in messages "Attachment not found")

810:
that's a other issue, you need to recount by yourself. Merging this one

Issue #3098, opened 2015-05-17
[K4.0] Crypsis: The Plugin Captcha - ReCaptcha (version 2.0) does not work with Kunena


rich20:
http://www.kunena.org/forum/kunena-4-0-support-reporting-defects/133648-recaptcha-not-working-beta3
The code is not accepted: 
Error
Incorrect captcha. Please try again.
But the code is correct. For the Joomla registration the captcha works.

Ruud68:
@xillibit fix works for me if I remove the if ( $template->getTemplateDetails()->kversion > 4.0) check :)

xillibit:
Do-you use crypsis or blue eagle template ? This check is to make it working with crypsis

Ruud68:
crypsis, but I think the issue is that the value for the check is not yet in b3. I have only patched this single file, not all the changes (waiting for rc1) :)

xillibit:
It seems that a previous change that i had done has been forgot, will look in the evening

xillibit:
Try now i have updated it

PR #3106, opened 2015-05-19
[K4.0] Crypsis: The Plugin Captcha - ReCaptcha (version 2.0) does not work with Kunena #3098

xillibit:


rich20:
Error
Incorrect captcha. Please try again.

xillibit:
@rich20 : you can try now, i have updated it

810:
Fatal Error: Can't use method return value in write context in components/com_kunena/controllers/topic.php on line 346

810:
I Get a notice on phpstorm  that's only working on php >= 5.5

xillibit:
try now you shouldn't have the fatal error

810:
- [x] When reply with captcha
- [ ] When reply without captcha - error: Fatal Error: Call to a member function getError() on a non-object in components/com_kunena/controllers/topic.php on line 360

rich20:
I've additional problems. The captcha appears fully ineffective.
With Chromium you need only mark checkbox and then submit.
bildschirmfoto vom 2015-05-20 17 55 36
With Firefox it does not matter if you choose wrong images. You can submit.
bildschirmfoto vom 2015-05-20 18 05 40

xillibit:
> When reply without captcha - error: Fatal Error: Call to a member function getError() on a non-object in components/com_kunena/controllers/topic.php on line 360

I will remove that part for now, but i need to find a way to display the error message if needed

Ruud68:
Hi, just added the changes to the files in this commit manually on Kunena 4.0 RC1 and now the captcha 2.0 works > both on new reply as on new topic! :+1:

xillibit:
Thanks to confirm that it works

Issue #3095, opened 2015-05-15
[K4.0] Crypsis: 3 party plugins don't showed on new topic


810:
KunenaBbCodeEditorButton not working

xillibit:
Which third parties plugins aren't showed on new topic ?

810:
like tex plugin, or my own plugins

Looks like onKunenaBbcodeEditorInit don't work

xillibit:
This one needs a couple of hours to be totally working

PR #3134, opened 2015-05-25
[K4.0] Crypsis: 3 party plugins don't showed on new topic #3095 (first try)

xillibit:


xillibit:
Should be working now

810:
not yet

xillibit:
Why there is a big space in the middle ? Do-you have enabled plugins for kunena bbcode editor ?

810:
Yes, I have enabled my plugins. Also the bbcode for code is wrong

PR #3135, opened 2015-05-29
[k4.0] Crypsis hide more info for guests posts

810:


Ruud68:
Hi, unfortunately it still shows the info.
when I change <?php if (!$user->guest) : ?>
to <?php if (!$user) : ?>
it does work

810:
updated the code, should work now

Ruud68:
works :)
with this code in the same file you can also make the rank image (no added value for a guest?) also hidden for guests
        <?php if (!empty($rankImage) && ($user->userid > 1)) : ?>
        <li>
                <?php echo $rankImage; ?>
        </li>
        <?php endif; ?>

810:
I looked at it. But for ux i think we need have a guest rank. Looks cleaner for me.

Ruud68:
:) cool with me :)

Issue #3133, opened 2015-05-25
[K4.0] Post ordering: When global setting is Newest Post First, guests continue to see Oldest Post First


rich20:
http://www.kunena.org/forum/kunena-4-0-support-reporting-defects/133723-post-ordering

rich20:
- [x] Crypsis - Success
- [x] Blue Eagle - Success

PR #3144, opened 2015-05-30
[K4.0] Post ordering: When global setting is Newest Post First, guests

xillibit:
continue to see Oldest Post First #3133


rich20:
- [x] Crypsis - Success
- [x] Blue Eagle - Success

PR #3157, opened 2015-06-02
Replacement of caret.js/atwho.js by using Boostrap typeahead in search/userlist

xillibit:


810:
Do we still need the files? or can you removed it also

xillibit:
This PR is the first part, i need to do the same change on quick reply/message textarea because it use caret.js/atwho.js to show smilies list in the github way

xillibit:
@rich20 or @810  : is-it working for you ?

rich20:
Previously has been loaded: caret.js, atwho.js and bootstrap-min.js
Now only bootstrap-min.js
is it what you mean?

xillibit:
Now on search and userlist the auto-suggest for username is using bootstrap ahead instead caret.js, atwho.js, so just check that auto-suggest work

rich20:
auto.suggest dos not work for me

xillibit:
For me ti works : http://screencast.com/t/HojiACHGRDo

810:
it works only when login

xillibit:
It has been always like that even in blue eagle

rich20:
> it works only when login

You're right. I had tested as guest.

810:
ok then its fine, you can start part2

PR #3166, opened 2015-06-04
Add bootstrap 3 template

810:


xillibit:
Are-you sure that build works ? Because i had to update just the yuicompressor to 2.4.8 but it has failed in yui-compressor-ant-task-0.5.jar

810:
I have fixed the issue on edit.js

Line 142 and 143

xillibit:
During build process do-we create the crypsis b3 variation into a separate package like for languages ? Because you have to it will available in download as separate from Kunena 4.0

810:
yeah we need to fix the template installer first. And then we need to create package on the build

Issue #3165, opened 2015-06-03
[K4.0] Template Manager : You can't install templates


rich20:
http://www.kunena.org/forum/kunena-4-0-support-reporting-defects/133792-back-end-i-can-t-install-template-when-upload-zip-file

I've tested it. With Kunena 3.0 you can install Crypsis without poblem.

With Kunena 4.0 it gives the message:
Notice
Missing file to extract

xillibit:
> I've tested it. With Kunena 3.0 you can install Crypsis without poblem.

Do-you have tested on same Joomla! version with k3.0 and k4.0 ?

rich20:
Yes, tested with Joomla 3.4.1 and J3.4.2dev. Also with Joomla 2.5.
With Kunena 3.0 it works everywhere, but not with Kunena 4.0

PR #3168, opened 2015-06-04
 [K4.0] Template Manager : You can't install templates #3165

xillibit:


rich20:
- [x] Success - works now with J 3.4 as well as J 2.5

Issue #3173, opened 2015-06-05
[K4.0] Crypsis : When ReCaptcha plugin is enabled registered users can't write


rich20:
The problem exists, when ReCaptcha plugin is enabled and CAPTCHA Challenge for Guests = Yes. Registered users can't write. Even if the correct code is entered.

Ruud68:
Hi @rich20 , is this the bug for this report? http://www.kunena.org/forum/kunena-4-0-support-reporting-defects/133757-bug-registered-user-cannot-post-in-kunena

rich20:
In principle it's the same. With your workaround only users and guests can write. But not admins or moderators.

810:
@rich20 could you check the fix. For me it worked.
- Plugin enabled and with keys
- Enabled CAPTCHA Challenge for Guests
- CAPTCHA Challenge for Users = 0
- kunena keys empty

- [x] Reply guests
- [x] Reply users
- [x] Reply admin
- [x] New Topic guests
- [x] New Topic users
- [x] New Topic admin

rich20:
- [x] Reply guests
- [x] New Topic guests
- [ ] Reply users
- [ ] Reply admin
- [ ] New Topic users
- [ ] New Topic admin

This is the last version which works for me. https://github.com/Kunena/Kunena-Forum/pull/3105 All later versions have this error.

xillibit:
Let me look to this one

PR #3184, opened 2015-06-07
[K4.0] Crypsis : When ReCaptcha plugin is enabled registered users can't write #3173 (Issue with mods/admins)

xillibit:


rich20:
- [x] Yes, now works correct.

PR #3226, opened 2015-06-15
 [K4.1] Libraries: Code Improvements

810:


libraries/kunena/template/template.php
```diff
@@ -1043,7 +1237,7 @@ public static function getInstance($name = null)
 		$app = JFactory::getApplication();
 		if (!$name)
 		{
-			$name = JRequest::getString ( 'kunena_template', KunenaFactory::getConfig()->template, 'COOKIE' );
+			$name = JFactory::getApplication()->input->getString ( 'kunena_template', KunenaFactory::getConfig()->template, 'COOKIE' );
```
xillibit: When it's available, you should use $app->input...

810: yeah saw that to late.

Issue #3228, opened 2015-06-15
[K4.0] Ban: delete all posts isnt working 


810:
I saw this issue since K3.0.6/K3.0.7.

When you ban a spammer, and select the option delete all posts.

Then its deleted. But you can still see the topic in the category, but delete. But you can't delete it fully.

xillibit:
> But you can still see the topic in the category, but delete.

You mean deleted ?

810:
yes, but you can't remove it totally.

PR #3231, opened 2015-06-16
 [K4.0] Ban: delete all posts isn't working #3228

xillibit:


810:
could you change it to global moderator

xillibit:
I don't think that i can do it

810:
Then its ok, there is another check its global. I will merge this one.

PR #3238, opened 2015-06-20
Add settings for auto-subscribe cats for new users and uncomment it

xillibit:


810:
@rich20 could you test this one

rich20:
It seems to be missing something. I can't find these option.

rich20:
Ok, I found it. When subscribe is enabled and you create a new user, it gives this error:

Save failed with the following error: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near ''' AS user_id, c.id as category_id, 1 FROM prefix_kunena_categories AS c ' at line 2 SQL=INSERT INTO prefix_kunena_user_categories (user_id,category_id,subscribed) SELECT {'' AS user_id, c.id as category_id, 1 FROM prefix_kunena_categories AS c LEFT JOIN prefix_kunena_user_categories AS s ON c.id=s.category_id AND s.user_id={'' WHERE c.parent_id>0 AND c.id IN (25) AND s.user_id IS NULL

rich20:
@810 with your edit {$db->quote($user->userid)} in line 219 and 221 the SQL error is gone. 

What does it mean "Subscribe users to all categories" ? All categories or all manually entered category IDs?

It only works, if you entered the IDs and additional both options set to Yes. It needs all 3 options at the same time.

sozzled:
It's actually not a good idea to have "autosubscribe" new users as a standard feature of the Kunena core software.  In some parts of the world, this is illegal to do this.

xillibit:
> It's actually not a good idea to have "autosubscribe" new users as a standard feature of the Kunena core software. In some parts of the world, this is illegal to do this.

By default it's disabled

sozzled:
It doesn't matter that it's disabled by default.  The fact that this exists as a configurable option in the Kunena core is no excuse in a court of law.  It might be better to include a feature such as this as a separately installed add-on.  I have given you my advice; if you want to ignore it then it's entirely a matter for you to prepare for any legal consequences that may arise.

In short, providing site owners with a mechanism to automatically opt-in their users to subsciptions when they join a website (without informing them of those condition) is in violation of anti-spam and privacy legislation in many countries around the world.

If this plan goes ahead, it will be challenged vigorously and may result in Kunena being removed from the JED.

Jiminimonka:
Michael, stop being a dick.

xillibit:
> It doesn't matter that it's disabled by default. The fact that this exists as a configurable option in the Kunena core is no excuse in a court of law.

I'am not a jurist but i'am interessed to know if it's the responsability of site owner or of software itself ?

Ruud68:
Hi, not a legal person either, but I think a warning (as in description for the configuration option) would be enough.
I use acymailing and they also have the option to autosubscribe on registration etc. no warnings and they are still on the JED.

What would be nice (not sure if this is a legal requirement) is to have an opt-out in the mails sent: stop receiving mail from forum. For news letters (here in the NL) that is a legal requirement.

810:
Just checked the legal rules. 

Auto subscription is allowed for B2B (business to business) and not allowed for B2C (business to customers)

So we can include the feature, and we can display a message its only allowed for B2B.

sozzled:
It is **_imperative**_ that users are able to opt-out from being "automatically" subscribed.  If an opt-out feature is not included, then this would not be legal.  This requirement is in APP7 of the Australian Privacy Principles (http://www.oaic.gov.au/privacy/privacy-resources/privacy-guides/australian-privacy-principles-and-national-privacy-principles-comparison-guide) and other jurisdictions have similar requirements as well.  The Kunena product must contain a strong disclaimer that the use of this feature **_may**_ conflict with local or international covenants on privacy.

Ruud68:
@sozzled does it also say something about where they can opt-out? In the received email via a link in eg the bottom or is going to the website, logging in, and disabling it sufficient?

sozzled:
The opt out must allow the subscriber to cancel **_all**_ email subscriptions from the website.  For example, if there are 150 categories and a user is automatically subscribed to all of them, it is insufficient to have to "unsubscribe" from each category, individually, every time an email is received.

Quite simply, this is a legal minefield that you guys are stepping into.  It is not a valid legal excuse to say that you did not know how the feature would be used (or how it could be used) unless a comprehensive legal disclaimer is included with **_every**_ installation kit of _Kunena_ where this feature is included.  If _Kunena_ does not include a comprehensive terms and conditions document, that fully explains that all responsibility is on the **_user**_ of the software and not on the **_developer**_ of the software, then the developer can be enjoined in legal and/or criminal proceedings brought against any user of the software.  Sorry, that's **_exactly**_ how the law operates.

You know how this feature can be used and abused.  The legal principle of **_due diligence**_ applies.  To say you did not know and did not take diligent steps to forewarn users about your and their legal obligations is inexcusable.  You know because I told you about this nearly 6 weeks ago.

As I intimated earlier in this topic, this is something you really need to get expert legal advice upon.  I, for my part, disclaim, disavow and abjure all further responsibility from the distribution of software that includes this feature.  I will not be held accountable for any civil litigation or criminal action brought against any individual or all people who are, currently or formerly, associated with the Kunena project.

If you go ahead with this idea, it is entirely your responsibility to defend and indemnify yourselves against any and all legal (or other) consequences that may ensue.

And, by the way, under the _Australian Trade Practices Act_, B2B emails are not preremptorily excused by the ambit of this feature unless contractual arrangements between two businesses make it clear that this feature is part of the usual means of communication that is employed between those two businesses.  So, no, B2B is not "OK" as Jelle would try to make everyone believe.  That's pure fiction.

Issue #3241, opened 2015-06-20
Parameter kunena_template does not work clearly


JProof:
File:\libraries\kunena\template\template.php
Line:~1000

there are an mistake between template-name and template-Classname
template names should be lower caser ?
tempateClassName is written in CamelCase?

so, the request parameter should be in Camelcase and template-name which is triggering the file-dir in lower case?

...i.e. solution

/**
	 * Returns the global KunenaTemplate object, only creating it if it doesn't already exist.
	 *
	 * @access        public
	 *
	 * @param        int $name Template name or null for default/selected template in your configuration
	 *
	 * @return        KunenaTemplate        The template object.
	 * @since         1.6
	 */
	public static function getInstance($name = null)
	{

		$app = JFactory::getApplication();
		if(!$name){

			$name = JFactory::getApplication()->input->cookie->getString('kunena_template', KunenaFactory::getConfig()->template);

		}

		$name = KunenaPath::clean($name);
		if(empty(self::$_instances[$name])){
			// Find overridden template class (use $templatename to avoid creating new objects if the template doesn't exist)
			$templatename = $name;
			// directories in lower case
			$relTmplDirName = strtolower($templatename);
			// classnames perhaps in CamelCase
			$classname = "KunenaTemplate{$templatename}";

			if(!is_file(KPATH_SITE . "/template/{$relTmplDirName}/template.xml") && !is_file(KPATH_SITE . "/template/{$relTmplDirName}/config.xml")){
				// If template xml doesn't exist, raise warning and use blue eagle instead
				$file         = JPATH_THEMES . "/{$app->getTemplate()}/html/com_kunena/template.php";
				$templatename = 'blue_eagle';
				$classname    = "KunenaTemplate{$templatename}";

				if(is_dir(KPATH_SITE . "/template/{$relTmplDirName}")){
					KunenaError::warning(JText::sprintf('COM_KUNENA_LIB_TEMPLATE_NOTICE_INCOMPATIBLE', $name, $templatename));
				}
			}

			if(!class_exists($classname) && $app->isSite()){

				$file = KPATH_SITE . "/template/{$relTmplDirName}/template.php";

				if(!is_file($file)){
					$classname = "KunenaTemplateBlue_Eagle";
					$file      = KPATH_SITE . "/template/blue_eagle/template.php";
				}

				if(is_file($file)){
					require_once $file;
				}
			}
			if(class_exists($classname)){
[...]

PR #3243, opened 2015-06-20
Parameter kunena_template does not work clearly #3241

810:


xillibit:
You have changed too much lines, i don't see what is the change

810:
the changes are on line 1062,1071,1075,1082

the only change is :

$relTmplDirName = strtolower($templatename); 

And then call this instead of $templatename

xillibit:
I have tried it and i it works

PR #3260, opened 2015-06-25
Check during install if fileinfo is enabled to use FINFO class

xillibit:


810:
What if the function is disabled, do you get a error, or is the installation stopping?

xillibit:
Kunena can't be installed and you the error message: "Required PHP extension '%s' is missing. Please install it into your system." Like for gd, json, dom, pcre, SimpleXML

Issue #3310, opened 2015-07-22
Support PDO Driver


acs-ferreira:
Since 3.4 PDO driver is the new reference for us when deploying Joomla.
Is PDO support on the Roadmap ?

Thanks.

xillibit:
Are-you talking about that : https://github.com/joomla/joomla-cms/pull/6175 ?

acs-ferreira:
No, when upgrading Kunena, we've a message ````PDO is not supported, please switch to MySQL````

xillibit:
The check in install part check database settings against mysql or mysqli, i will modify it to check against PDO too

xillibit:
How to reproduce the issue ? In your php.ini what extensions do-you have enabled ?

acs-ferreira:
# To reproduce
1. Install Kunena 4.0 with MySQL(i) as driver on on Joomla Configuration
2. Swicth to PDO on Joomla Configuration
3. Try to update from 4.0 to 4.0.3

# php.ini
````
--enable-fpm
--enable-opcache
--enable-sockets
--enable-pcntl
--with-readline
--enable-ftp
--enable-shmop
--with-snmp
--enable-soap
--enable-zip
--enable-calendar
--with-openssl
--with-curl
--with-mysql
--with-mysqli
--with-pgsql
--with-pdo-mysql
--with-pdo-sqlite
--with-pdo-pgsql
--with-unixODBC
--with-pdo-odbc
--with-bz2
--with-zlib
--with-gd
--enable-gd-native-ttf
--with-freetype-dir=/usr
--with-jpeg-dir=/usr
--with-png-dir=/usr
--enable-exif
--with-xsl
--enable-xslt
--with-xslt-sablot
--with-mcrypt
--enable-mbstring
--enable-intl
--enable-bcmath
--with-gettext
--with-tidy
--with-imap
--with-imap-ssl
--with-kerberos
````

810:
Ok we have added pdo support from the next kunena version 4.0.11

PR #3312, opened 2015-07-23
 Support PDO Driver #3310

xillibit:


810:
change also:
https://github.com/xillibit/Kunena-forum/blob/k4.0_bugs2/install.php#L124

810:
Ok we can now support pdo

Issue #3328, opened 2015-08-01
[K4.0] Images not showing in RSS feed	


rich20:
http://www.kunena.org/forum/kunena-4-0-support-reporting-defects/134352-images-not-showing-in-rss-feed

It's the same on k.org. It shows only the code  [File Attachment: imagename.jpg]

sozzled:
Images used to show in RSS feeds from http://kunena.org/forum when the site was using K 3.0.8

If it doesn't work now (at K.org) is it something that has changed since the site was updated to K 4.0

810:
merged

PR #3329, opened 2015-08-01
Images not showing in RSS feed #3328

810:


rich20:
- [x] Success

xillibit:
Check in Jomsocial activity streams if it's still wokr like before

810:
I have added insert attach.

But we need to look in the jomsocial plugin. Old bug (no issue for it).  When you disables show image/attach for guests, the image/attach is always added on the stream, because the jomsocial activity is just plain text.

We can't change this behavior.

Issue #3318, opened 2015-07-27
[K4.0] Max Links in message will divided by three if you insert a image link


rich20:
You have to many links in your message, please decrease it!
Example:
Max Links in message = 6
You can put in the message either 2 image links - or 1 image and 1 normal link
Max Links in message = 9
You can put in the message either 3 image links - or 2 images and 1 normal link, and so on

810:
Don't have this issue on K4.0.4.dev2

Max Links in message = 6
Can upload 4 links and 2 inserted images from attachments


Or are you use embed images

rich20:
Images via url such as [img]address.png[/img]

xillibit:
> Max Links in message = 6
> You can put in the message either 2 image links - or 1 image and 1 normal link

I can put six embed images via url such as [img]address.png[/img], so i'am not able to reproduce the issue

rich20:
You've tried as admin? A admin has no limit. Please try it as registered user.

xillibit:
I have tested as registered user and i'am to reprodue the issue, but if you insert two images (with [img]address.png[/img]) it counts 6 because there 4 http and 2 href

PR #3334, opened 2015-08-06
 [K4.0] Max Links in message will divided by three if you insert a image

xillibit:
link #3318


810:
The function is twice in the file. So you need to edit it to

xillibit:
The method checkMaxLinks() is only present one time in the file, i had defined this method to avoid duplicate code

810:
now you can always unlimited embed images

xillibit:
Try now i may have done a mistake before

810:
again divided by three

xillibit:
try now

rich20:
- [x] Success

PR #3335, opened 2015-08-06
 Crypsis: Edit poll is not possible (it missing the poll symbol in the

xillibit:
editor) #3332

PR #3336, opened 2015-08-08
Crypsis: Edit poll is not possible (it missing the poll symbol) #3335

xillibit:


810:
i see still not the icon with IE

xillibit:
Try now

810:
+1

Issue #3326, opened 2015-07-31
Feature Request: Kunena Menu Option - All Unread Items


GJSchaller:
Similar to Recent Topics menu item, but instead a list of all topics that have unread content for that user.

810:
duplicated issue, closing this one

PR #3342, opened 2015-08-10
Feature Request: Kunena Menu Option - All Unread Items #3326

810:


rich20:
- [ ] It shows the same as the menu item My Topics.

PR #3386, opened 2015-08-24
Fix error on default.js

810:
real fix line 335


xillibit:
Did-you have reproduced the error which have the user ?

810:
No, only by checking the code. First code is return false, then there is a if function, then return true.

So the error should be correct.
I will check later, if its works

810:
yes, reproduced, and the code is working correct now

PR #3388, opened 2015-08-27
Crypsis: Seo improvement

810:


xillibit:
Why do-you need to add nofollow here ?

810:
Because google indexed reply. So you get duplicated meta and title. This is the only linking to reply without nofollow.

PR #3439, opened 2015-09-10
Fix installer issues

810:


xillibit:
->query() is deprecated you should use ->execute() instead it's the reason why it was replaced, there is a typo in the first line 236

810:
Oh didn't see the typo. You can create a PR for it.

xillibit:
Ok, will do a new PR for it

PR #3446, opened 2015-09-18
Wrong path to display attachments in backend files manager

xillibit:


810:
could you merge develop branch. because now its deleting the check

PR #3455, opened 2015-09-24
Need to pass files as raw to be avoid to be filtered and cause null result

xillibit:
This is caused by PR7814 in Joomla repo


810:
+1

PR #3456, opened 2015-09-24
File upload result in COM_KUNENA_UPLOAD_ERROR_FILE_E​XCEED_LIMIT_IN_CONFIGURATION

810:


libraries/kunena/upload/upload.php
```diff
@@ -569,7 +569,7 @@ public function upload($fileInput, $destination, $type = 'attachment')
 			}
 		}
 
-		if (!$this->checkFileSize($file->size, $file->isAvatar))
```
xillibit, line 572: By looking quickly `$file->isAvatar` was just undefined in one case....

xillibit:
@jelle : I don't remember what was the purpose of this fix ?  Can-you give the link of Kunena topic where the issue is described ?

810:
see: http://www.kunena.org/forum/k-4-0-support-archive/134072-file-upload-result-in-com-kunena-upload-error-file-exceed-limit-in-configuration?start=10

PR #3465, opened 2015-09-28
Feature: Social Share

810:


Ruud68:
Hi, really like this functionality :)
Would be great if the icon-theme would also be a config option.
That way the site / forum owner can choose between:

```
jssocials-theme-flat.css - flat theme
jssocials-theme-classic.css - classical theme with raised buttons
jssocials-theme-minima.css - minimalistic theme with logos instead of buttons
jssocials-theme-plain.css - monochromatic theme
```

The themes are part of the js-socials package :)

810:
Updated the code

Issue #3483, opened 2015-10-11
Poll icon always showed


810:
When you create a new topic,

You see the poll icon, but when you select a other category, then change the category back. The icon is gone.

So only first load is wrong.

PR #3484, opened 2015-10-11
Poll icon always showed #3483

810:


rich20:
The poll icon is available also if you reply. No matter which category.

810:
also after the fix?

rich20:
Yes, I had tested it with this fix.

xillibit:
I can do a PR for it in the evening because i have found a way to fix it

810:
fixed by other pr

PR #3489, opened 2015-10-17
Added EasySocial plugin for Kunena

imarklee:


810:
Could you change the branch to kunena:K5.0

PR #3492, opened 2015-10-17
Poll icon always showed #3483

xillibit:


xillibit:
@rich20 : do-you have tested it ? Is-it work like it should ?

810:
+1

PR #3493, opened 2015-10-18
Easyprofile integration plugin Kunena

Ruud68:
Hi, no experience with github so hope this works :)


810:
Hi Ruud, thnx, but this isn't correct. 

Try to do it again, but you merged master branch before commit your changes. Now try to do the same, only with K5.0 branch

Issue #3404, opened 2015-09-04
New email template


810:
- Modern template
- (Legal) add unsubscribe.   (unsubscribe topic + unsubscribe options)

PR #3496, opened 2015-10-25
New responsive email template

810:
https://github.com/Kunena/Kunena-Forum/issues/3404


GJSchaller:
I noticed in the current notifications, emojis don't have a proper URL:

(img src="http:///media/kunena/emoticons/sad.png")

instead of a full URL - this may need to be adjusted in the new template.

Issue #3466, opened 2015-09-29
Check attachments size on upload take care of avatar setting instead of images/files settings


xillibit:
See : http://www.kunena.org/forum/k-4-0-general-questions/134726-update-4-0-6#168954

810:
avatar size = 290kb

test 1:
avatar size setting = 350kb
upload image size setting = 1500kb

Result:
Avatar uploaded correct

test 2:
avatar size setting = 150kb
upload image size setting = 1500kb

Result:
Avatar not uploaded, showed correct error

test 3:
avatar size setting = 350kb
upload image size setting = 150kb

Result:
Avatar uploaded correct.

test 4:
avatar size setting = 150kb
upload image size setting = 150kb

Result:
Avatar not uploaded, showed correct error.

xillibit:
You have done the test in reverse, you need to check the upload of attachments images

Just re-do the same test but by uploading attachments images

810:
Should be fixed now, can someone confirm

rich20:
-  [ ] Failed - Image upload needs both settings. 

Larger images can be uploaded only if you set both options higher
Maximum Image File Size 
and 
Maximum File Size

rich20:
It's only with Blue Eagle. With Crypsis works the setting corect.

810:
lasted build:

1) template: blue eagle
2) max avatar size = 2400kb , max image size = 150kb, max file = 120kb
3) Avatar 380kb
4) Result: Avatar uploaded successfully.
Your profile has been saved.

rich20:
> You have done the test in reverse, you need to check the upload of attachments images
> Just re-do the same test but by uploading attachments images

I've tested also this. Or did he meant avatars? Now is the problem with upload images in messages. That's what I meant. With avatars it works.

810:
yes, this issue is only with the avatar upload.

810:
fixed

PR #3510, opened 2015-11-04
Check attachments size on upload take care of avatar setting instead of images/files settings #3466

810:
### Blue Eagle
- [x] Upload avatar
- [x] Show error upload avatar when reached limit
- [x] Upload image
- [x] Show error when upload image limit is reached
- [x] Upload File
- [x] Show error when file upload limit is reached
- [x] Split error message
- [x] Split limit attachments
### Crypsis/Crypsisb3 (still need tests)
- [x] Upload avatar
- [ ] Show error upload avatar when reached limit - Failed show correct error but standard default avatar is now set as default
- [x] Upload image
- [x] Show error when upload image limit is reached
- [x] Upload File
- [x] Show error when file upload limit is reached
- [x] Split error message
- [x] Split limit attachments


rich20:
In principle it works all fine. But if you try upload big attachments/images with Crypsis, you can send the message before the check is finished. The message will saved without error message and attachments.

Issue #3517, opened 2015-11-11
Crypsis: Target topic does not change to another category


rich20:
Move topic or move message. If you want move to another category, the target topics stays from the first category.
Example: Move from "Welcome Mat" to "Suggestion Box". You see the target topics from Welcome Mat (image 1). Now I switch to Suggestion Box. The target topics are the same (image 2)
![bildschirmfoto vom 2015-11-11 19 37 18](https://cloud.githubusercontent.com/assets/888847/11099978/b6c38ed8-88ae-11e5-9517-230e1f57c7aa.png)
![bildschirmfoto vom 2015-11-11 19 39 22](https://cloud.githubusercontent.com/assets/888847/11099984/c1a72652-88ae-11e5-817d-ee300853beb3.png)

810:
Looks like we miss some js from media\kunena\js\default.js

@xillibit could you convert this to jquery on the crypsis

810:
merged

PR #3524, opened 2015-11-14
Crypsis: Target topic does not change to another category #3517

xillibit:


810:
also do crypsisb3

rich20:
-  [ ] Failed - no reaction

> also do crypsisb3

it's the same

810:
Is the fix working rich. because you selected the box. but you entered Failed

Also is "it's the same" working or not

rich20:
Sorry, I had automatically inserted the x. My mistake and I have it changed. It's the same, both does not work.

xillibit:
Try now

810:
Almost, you need to open the target selectbox 2x to see the right categories

rich20:
- [x] Success

xillibit:
That's the same behaviour that on blue eagle

810:
ok, its the delay, can we improve that more, or is it to hard to find where the long delay is.

xillibit:
It does an ajax request so it's maybe the reason why you have a dalay

810:
ok, I merge this one for now.

Issue #3525, opened 2015-11-14
Crypsis: You can reply anonymously but not create anonymously a new topic with the menu item "New Topic"


rich20:
You must to go the detour "Index -> Category -> New Topic"

810:
Looks we like missing the js also for this. @xillibit could you fix that.

810:
merged

PR #3526, opened 2015-11-14
Crypsis: You can reply anonymously but not create anonymously a new topic with the menu item "New Topic" #3525

xillibit:


810:
+1

rich20:
This option does not work. Here you can see the difference between Crypsis and Blue Eagle:

bildschirmfoto vom 2015-12-11 18 24 44

Issue #1109, opened 2012-09-21
Menu item "Page Display" options not honored


sozzled:
See http://www.kunena.org/forum/K-2-0-General-Questions/125040-changing-browser-title

810:
fixed on next version

PR #3528, opened 2015-11-15
Menu item "Page Display" options not honored #1109

810:
First HMVC


810:
@rich20 could you check this one. I have no error on Recent posts (logout and login)

Issue #3533, opened 2015-11-16
quick-reply and require email error


Ruud68:
Hi,
yesterday I noticed that the fix for adding the email field to the quick reply is not fully working :(
It works correct for quests, but when logged in the email is also required, this should not be.

banging my head against the wall for having tested this to quickly.

anyway, the fix is in components/com_kunena/template/crypsis/layouts/message/edit/quickreply.php

replace

```
<?php if ($config->askemail): ?>
```

with

```
<?php if ($config->askemail && !$me->exists()): ?>
```

xillibit:
At which line did-you seen that ? Because it already check if the user doesn't exist

Ruud68:
Hi, as you can see here, it does check for user logged in, but the display of the emailfield is outside that if statement...

```
                        <?php if (!$me->exists()) : ?>
                                <div class="controls">
                                        <label>
                                                <?php echo JText::_('COM_KUNENA_GEN_NAME'); ?>:
                                        </label>
                                        <input type="text" name="authorname" class="span12" maxlength="35" placeholder="<?php echo JText::_('COM_KUNE$
                                </div>
                        <?php endif; ?>

                        <?php if ($config->askemail): ?>
                                <div class="controls">
                                        <?php echo $config->showemail == '0' ? JText::_('COM_KUNENA_POST_EMAIL_NEVER') : JText::_('COM_KUNENA_POST_EM$
                                        <input type="text" id="email" name="email" placeholder="<?php echo JText::_('COM_KUNENA_TOPIC_EDIT_PLACEHOLDE$
                                </div>
                        <?php endif; ?>
'''
```

xillibit:
I see now the line that you are mentionning, i will fix it later

810:
I had removed it because in the description its said also for registered users.

"Require an e-mail address when users or visitors make a post" That's why I removed it

Ruud68:
hi, I was/am under the assumption that a user always has an email address (as part of the registration process), so asking for the email address is only applicable to guests. Correct?
If so, the above mentioned description should be changed to: "Require an e-mail address when visitors make a post"

xillibit:
We know email for registered users so yes the sentence above isn't correct i think

810:
ok merged.

PR #3539, opened 2015-11-17
quick-reply and require email error #3533

xillibit:


810:
could you update the language string, so remove "users.."

Issue #3575, opened 2015-12-02
[K5.0] Configuration -> (tab) Frontend  -> Show Moderators in Category Index : It shows Undefined index


rich20:
Notice: Undefined index: listcat_show_moderators in components/com_kunena/template/config/default.php on line 282

810:
merged

PR #3576, opened 2015-12-02
[K5.0] Configuration -> (tab) Frontend -> Show Moderators in Category Index : It shows Undefined index #3575

810:


rich20:
- [x] Success

Issue #3585, opened 2015-12-05
Attachments : With the button "Insert All" get all attachments the value "null"


rich20:
[attachment=null]schlawiner.jpg[/attachment] [attachment=null]pizza_3.jpg[/attachment]

810:
fixed and merged

PR #3591, opened 2015-12-09
Attachments : With the button "Insert All" get all attachments the value "null" #3585

810:


xillibit:
Avoid to do in the same pull request the style change and the fix because i can't see sometime where is exactly the fix

810:
phpstorm does this for me automatically when I save a file.

The change is on line 92

Issue #3623, opened 2016-01-12
Category moderator can't upload attachments in his category if is set "Allow Uploads = Moderator"


rich20:
https://www.kunena.org/forum/k-4-0-general-questions/135830-upload-file-as-a-moderator-to-a-specific-category
This problem exists at both settings
Allow Image Uploads = Moderator
Allow File Uploads = Moderator

Message:  You do not have permissions to access this page.

PR #3627, opened 2016-01-13
Category moderator can't upload attachments in his category if is set "Allow Uploads = Moderator" #3623

810:


rich20:
- [x] Success

Issue #3635, opened 2016-01-20
When Direct Component Access and Protect Attachments is active, Images will not appear in the backend


rich20:
https://www.kunena.org/forum/k-4-0-general-questions/135913-attachment-images-missing-in-back-end-after-migration
This problem exists if you choose both settings:
Protect Attachments = Yes
Direct Component Access = No

xillibit:
I think i have found the root of the issue, but i'am looking to do a fix for it

rich20:
- [ ] Failed

PR #3647, opened 2016-01-23
When Direct Component Access and Protect Attachments is active, Images

xillibit:
will not appear in the backend #3635

### DIFF
--- a/components/com_kunena/admin/template/joomla30/layouts/attachment/item/default.php
+++ b/components/com_kunena/admin/template/joomla30/layouts/attachment/item/default.php
@@ -12,11 +12,26 @@ defined ( '_JEXEC' ) or die ();
 
 /** @var KunenaAttachment $attachment */
 $attachment = $this->attachment;
+
+if ($attachment->protected)
+{
+	$url_href = $attachment->getUrl();
+}
+else
+{
+	$url_href = JUri::root() . $attachment->getUrl();
+}
+
 ?>
-<a href="<?php echo JUri::root() . $attachment->getUrl(); ?>" title="<?php echo $attachment->getFilename(); ?>">
-	<?php if ($attachment->isImage())
+<a href="<?php echo $url_href; ?>" title="<?php echo $attachment->getFilename(); ?>">
+	<?php
+	if ($attachment->isImage() && !$attachment->protected)
 	{
 		echo '<img src="' . JUri::root() . $attachment->getUrl(true) . ' " height="40" width="40" />';
+	}
+	elseif ($attachment->isImage() && $attachment->protected)
+	{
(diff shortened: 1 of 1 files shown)


rich20:
- [ ] Failed

xillibit:
What is the link that you have of thumb image with the fix ?

810:
http://www.rich20.at/test-5-487329874//test-5-487329874/index.php?option=com_kunena&view=attachment&id=28&thumb=1&format=raw

xillibit:
With this fix the link on thumbs for protected attachments are like that : /joomla3.3.6/index.php?option=com_kunena&view=attachment&id=1&thumb=1&format=raw

810:
i miss the administrator dir

rich20:
- [ ] Does not work without SEO (no thumbs and wrong address)

xillibit:
I hadn't checked without SEO, i will check with that

810:
@xillibit could you merge my pr

xillibit:
> @xillibit could you merge my pr

Do-i still need to merge your PR ?

810:
yes. Could you test it also

xillibit:
When you set Direct Component Access on No, it check the menu active item and if it doesn't find or find the menu item is wrong it display a 404 page. So in this case i don't know how to display attachments in backend.

810:
then load the icon

`echo '<i class="'. $icon .' icon-big"></i>';`

rich20:
> then load the icon
> 
> `echo '<i class="'. $icon .' icon-big"></i>';`

That solves not the problem with the wrong addresses.

> With this fix the link on thumbs for protected attachments are like that : /joomla3.3.6/index.php?option=com_kunena&view=attachment&id=1&thumb=1&format=raw

format=raw&itemid=xxxx
If the itemid is added at the end of the address it works for both (link to images and it shows thumbs). It works with and without SEO

xillibit:
If i add the &itemid=xxxx at the of url it doesn't work better

810:
maybe we should disable link on the backend on that, when its protected and sef is turned off.

xillibit:
Try now

rich20:
- [x] Direct Component = No - it shows only icons without link
- [x] Direct Component = Yes - the images will displayed correct
- [x] Direct Component = Yes - the attachments can be called using the browser
- [ ] It remain the problem with the double address such as 

website/joomla-test/joomla-test/index.php?option=com_kunena&view=attachment&id=3&format=raw

xillibit:
Try now

rich20:
- [x] Direct Component = No - it shows only icons without link
- [x] Direct Component = Yes - the images will displayed correct
- [x] Direct Component = Yes - the attachments can be called using the browser
- [x] It remain the problem with the double address

Issue #3690, opened 2016-02-22
BBCodes for maps should be hardcoded


rich20:
#### Steps to reproduce the issue

https://www.kunena.org/forum/kunena-support-reporting-defects/136268-wrong-map-bbcode-syntax-on-non-english-4-0-10
1. Use a different language as English
2. Editor - click on map symbol -> Map Settings -> Choose a translated type of map -> Add 
#### Expected result

`[map type=Terrain zoom=2][/map]`
#### Actual result

`[map type=Gelände zoom=2][/map]` (depending on the language translation)
#### System information (as much as possible)

Joomla version: All
Kunena version: K 4.0.10
Php version:
Database version:
#### Additional comments

**BBCode works only with english language but it shows the text from the ini file.**
Line 39 in "components/com_kunena/template/crypsis/layouts/topic/edit/editor/default.php". These 4 options should be hardcoded.
`JText::_('COM_KUNENA_EDITOR_MODAL_MAP_SETTINGS_HYBRID') ?></option><option type="ROADMAP"><?php echo JText::_('COM_KUNENA_EDITOR_MODAL_MAP_SETTINGS_ROADMAP') ?></option><option type="TERRAIN"><?php echo JText::_('COM_KUNENA_EDITOR_MODAL_MAP_SETTINGS_TERRAIN') ?></option><option type="SATELLITE"><?php echo JText::_('COM_KUNENA_EDITOR_MODAL_MAP_SETTINGS_SATELLITE') ?></option></select><br />`

**These strings are then unnecessary:**
`COM_KUNENA_EDITOR_MODAL_MAP_SETTINGS_HYBRID="Hybrid"
COM_KUNENA_EDITOR_MODAL_MAP_SETTINGS_ROADMAP="Roadmap"
COM_KUNENA_EDITOR_MODAL_MAP_SETTINGS_TERRAIN="Terrain"
COM_KUNENA_EDITOR_MODAL_MAP_SETTINGS_SATELLITE="Satellite"`
(components/com_kunena/language/en-GB/en-GB.kunena_tmpl_crypsis.ini lines 113 - 116)

810:
merged

PR #3691, opened 2016-02-23
 BBCodes for maps should be hardcoded #3690

xillibit:
Pull Request for Issue #3690 .


rich20:
- [x] It works, however the code is written with uppercase letters

`[map type=TERRAIN zoom=2]Paris[/map]`

810:
Is it only on Crypsis?

rich20:
Yes, these setting options are available only for Crypsis.

xillibit:
I have set type value in lower case

rich20:
- [x] Success

Issue #3696, opened 2016-02-24
Ban: force logout user


810:
#### Steps to reproduce the issue

We need to logout the user before we run the ban function, to reduce the sever memory, if the user is spamming.
#### Expected result

Logout user when do ban
#### Actual result

User can send spam in the mean time, results the server will be slow
#### System information (as much as possible)

Joomla version: All
Kunena version: All
Php version: All
Database version: All
#### Additional comments

None

810:
merged

PR #3698, opened 2016-02-24
 Ban: force logout user #3696

xillibit:
Pull Request for Issue #3696.

### DIFF
--- a/libraries/kunena/user/ban.php
+++ b/libraries/kunena/user/ban.php
@@ -546,6 +546,11 @@ class KunenaUserBan extends JObject
 			return false;
 		}
 
+		// Logout the user before to ban him.
+		$user = JFactory::getUser($this->userid);
+		$app = JFactory::getApplication ();
+		$app->logout((int) $this->userid);
+
 		if (!$this->id)
 		{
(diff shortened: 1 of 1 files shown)


libraries/kunena/user/ban.php
```diff
@@ -613,7 +619,6 @@ public function save($updateOnly = false)
 			if ($block)
```
810: then you can remove this check

xillibit:
@rich20 : can-you test it please ?

rich20:
I think, my test will not be helpful.
At ban from the page I can't see whether the user is logged out before he is banned. It's too quick.  I see logout and user is banned at the same time.

And at ban from the forum, user stay logged on.

810:
Yes, the user is still logged on. After banning

810:
With this code it worked:

```
$user = JFactory::getUser($this->userid);
$app = JFactory::getApplication ();
$app->logout((int) $this->userid);
```

xillibit:
Updated

810:
thats why the user isn't logged out right now.

PR #3701, opened 2016-02-26
Add utf8mb4 support

810:
Please test with new databases

1) first database any uft8 collation
2) second database any uft8mb4 collation

### DIFF
--- a/components/com_kunena/admin/install/install.xml
+++ b/components/com_kunena/admin/install/install.xml
@@ -4,7 +4,7 @@
 	<table name="kunena_aliases">
 		<field primary_key="yes"
 		       name="alias"
-		       type="varchar(255)"
+		       type="varchar(190)"
 		       null="0"/>
 		<field name="type"
 		       type="varchar(10)"
--- a/components/com_kunena/admin/install/schema.php
+++ b/components/com_kunena/admin/install/schema.php
@@ -748,9 +748,6 @@ class KunenaModelSchema extends JModelLegacy
 					if (!strstr($collation, 'utf8') && !strstr($collation, 'utf8mb4'))
 					{
 						$collation = 'utf8_general_ci';
-					}
-					if (!strstr($collation, 'utf8mb4'))
-					{
 						$str .= implode(",\n", $fields) . " ) DEFAULT CHARACTER SET utf8 COLLATE {$collation};";
 					}
 				break;
--- a/components/com_kunena/admin/install/schema.php
+++ b/components/com_kunena/admin/install/schema.php
@@ -740,6 +740,12 @@ class KunenaModelSchema extends JModelLegacy
 						}
 					}
 					$collation = $this->db->getCollation();
+
+					if (!strstr($collation, 'utf8') && !strstr($collation, 'utf8mb4'))
+					{
+						$collation = 'utf8_general_ci';
+					}
+
 					if (strstr($collation, 'utf8mb4'))
 					{
 						$str .= implode(",\n", $fields) . " ) DEFAULT CHARACTER SET utf8mb4 COLLATE {$collation};";
--- a/components/com_kunena/admin/install/schema.php
+++ b/components/com_kunena/admin/install/schema.php
@@ -740,12 +740,20 @@ class KunenaModelSchema extends JModelLegacy
 						}
 					}
 					$collation = $this->db->getCollation();
-					if (!strstr($collation, 'utf8'))
+					if (strstr($collation, 'utf8mb4'))
+					{
+						$collation = 'utf8mb4_general_ci';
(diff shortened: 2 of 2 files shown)


rich20:
##### Tested with database collation utf8mb4_general_ci
- [ ] New Joomla 3.4.8 installation created collation utf8_general_ci tables - Kunena can't be installed
- [ ] Upgrade to Joomla 3.5 several tables are changed now to utf8mb4_general_ci - Kunena can't be installed
- [x] New Joomla 3.5 installation created collation utf8mb4_general_ci - Kunena installation works (also without this fix). But the *_kunena_tables are utf8_general_ci

mahagr:
I think that the idea is in UTF8 to keep the collation what user has been set it to be to make search to work with users language. Right now it doesn't do that.

mahagr:
One thing.. My guess is that Joomla 3.5 always sets the default database character set, meaning that the code for setting table collation wouldn't be needed at all in it. Even Joomla 3.4 may have the same feature meaning that we could just remove the extra code and assume that collation and charset are already correct.

But even if its not true in J3.4, we could just test if J3.5 is present and have simpler logic for it.

BTW: The code which was changed was implemented for Joomla 1.5-1.7... So things may have changed since then.

rich20:
Joomla 3.5 beta 3 uses now utf8mb4_unicode_ci, regardless of whether the general database setting is utf8_general_ci.
The problem with Kunena exists only then, when general setting is utf8mb4_unicode_ci.
The installation stagnates after creation table *_kunena_version (state migrateDatabase)
table

810:
@rich20 try now

rich20:
##### Database with collation utf8mb4_unicode_ci
- [x] Kunena creates tables utf8mb4_unicode_ci - Success
- [x] New installation of Kunena - Success

##### Database with collation utf8_general_ci
- [ ] Kunena can't be installed - it stays on Preparing installation...

mahagr:
There's still the issue that the code always changes the collation where it should keep it...

810:
try now

rich20:
##### Database with collation utf8_general_ci
- [x] Kunena creates tables utf8_general_ci - Installation Success
- [x] Migrate from fb_tables - Success

##### Database with collation utf8mb4_unicode_ci
- [x] Kunena creates tables utf8mb4_unicode_ci - Installation Success
- [ ] Migrate from fb_tables - Failed (it's not possible to migrate if is collation utf8mb4)

mahagr:
Alright, I just checked the code and its much better. There's one issue though, you need to also make a case on collation not matching either utf8 -- in that case you should just use utf8 / utf8_general_ci instead..?

Regarding to migration from fb-tables, maybe migration should only work up to J3.4? I think that after that it might be better just to use importer component for migration...

810:
I don't mind if migration doesn't work on utf8mb4. Then you need to do migrate to kunena first on the old site. Or just use the utf8 database. Maybe we can improve it later, but it is not a high priority.

mahagr:
Alright, this latest version should work now.

I agree that migration isn't high priority. The only question that remains: should it be disabled in J3.5?

810:
the migration is now working again, only database is changed to utf8_general again. So you need to use admin tools to change kunena tables and the default collation

810:
or change the fb tables to utf8mb4 first, then do the migration. Then all is working normal.

810:
maybe we should also remove: https://github.com/Kunena/Kunena-Forum/commit/41ca2ad469ba9fe08edd2f730090f986932020f4#diff-dc620cdc97a9d08c39f676ef92baff4dL2158

```
$create = preg_replace('/(DEFAULT )?CHARSET=[\w\d]+/', '', $create);  
$create = preg_replace('/COLLATE [\w\d_]+/', '', $create);  
$create = preg_replace('/TYPE\s*=?/', 'ENGINE=', $create);  
```

Then maybe we can also support pdomysql

@mahagr what do you think about this. It should be handled by joomla. We should take the default settings

810:
ok fixed it on the K5 branch.

mahagr:
Does Joomla handle it in J3.4 as well? If it does, then we do not need this.

mahagr:
Related discussion in Joomla:

https://github.com/joomla/joomla-cms/issues/9423

richard67:
@mahagr 

> One thing.. My guess is that Joomla 3.5 always sets the default database character set

As far as I know, and I was a bit involved in doing the utf8mb4 stuff for Joomla! core, Joomla! core does not set the default character set of the whole database, except in the case that you specify an not existing database when doing a new installation and have privileges for the specified user on the specified server to create new databases, which is normally only done so on local testing environments.

In all other cases Joomla! specifies charset and collation for each single core table in the create table statements and so not depends on the default database collation.

ZerooCool:
#6572

PR #3756, opened 2016-03-28
Fix fatal error with php 7 : Cannot use 'String' as class name as it is reserved

xillibit:
Pull Request for Issue # .


810:
We should not add this, because there are more php7 fixes needed, lets finish k5.0

PR #3767, opened 2016-04-01
Check if title of topic or message contains URL to limit spam

xillibit:
Pull Request for Issue # .

### DIFF
--- a/src/components/com_kunena/controllers/topic.php
+++ b/src/components/com_kunena/controllers/topic.php
@@ -482,6 +482,16 @@ class KunenaControllerTopic extends KunenaController
 			}
 		}
 
+		$url_subject = $this->checkURLInSubject($message->subject);
+
+		if ($url_subject && $this->config->url_subject_topic)
+		{
+			$this->app->enqueueMessage(JText::_('COM_KUNENA_TOPIC_MESSAGES_ERROR_URL_IN_SUBJECT'), 'error');
+			$this->setRedirectBack();
+
+			return;
+		}
+
 		// Make sure that message has visible content (text, images or objects) to be shown.
 		$text = KunenaHtmlParser::parseBBCode($message->message);
 
@@ -497,7 +507,6 @@ class KunenaControllerTopic extends KunenaController
 
 			return;
 		}
-
 		$maxlinks = $this->checkMaxLinks($text, $topic);
 
 		if (!$maxlinks)
@@ -705,6 +714,16 @@ class KunenaControllerTopic extends KunenaController
 			}
 		}
 
+		$url_subject = $this->checkURLInSubject($message->subject);
+
+		if ($url_subject && $this->config->url_subject_topic)
+		{
+			$this->app->enqueueMessage(JText::_('COM_KUNENA_TOPIC_MESSAGES_ERROR_URL_IN_SUBJECT'), 'error');
+			$this->setRedirectBack();
+
+			return;
+		}
+
 		// Set topic icon if permitted
 		if ($this->config->topicicons && isset($fields['icon_id']) && $topic->authorise('edit', null, false))
 		{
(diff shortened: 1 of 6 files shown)


rich20:
- [ ] Two messages are displayed at the same time.
  Message 1 - You cannot put an URL in the subject of the topic or message.
  Message 2 - Your message has been successfully posted.
- [ ] The topic with Url will published

sozzled:
Wait, wait, wait ...

Hang on a minute:  are you guys saying that it would be **_wrong_** to create a topic on a Kunena forum that had a subject like "I cannot access http://www.kunena.org"?  Is that what you're saying?

Why is this change required?  Who asked for it?  Even worse, if someone creates a perfectly innocent topic with a URL in the subject, they will be banned for 48 hours!

This is not a good idea, in my opinion, and the code doesn't make allowances for perfectly reasonable cases where there is nothing [ethically/morally] wrong with including a URL in the subject of a forum topic.

xillibit:
> Hang on a minute: are you guys saying that it would be wrong to create a topic on a Kunena forum that had a subject like "I cannot access http://www.kunena.org"? Is that what you're saying?

Did-you have seen a lot of topics from non-spammers with URL in subject ? Half of messages posted by spammers has just URL in title, so it's the reason of this change

> Why is this change required? Who asked for it? Even worse, if someone creates a perfectly innocent topic with a URL in the subject, they will be banned for 48 hours!

Nobody is banned, i had added this string but the ban has been removed

> This is not a good idea, in my opinion, and the code doesn't make allowances for perfectly reasonable cases where there is nothing [ethically/morally] wrong with including a URL in the subject of a forum topic.

How do-you make difference between URLs morally correct and spam ?

sozzled:
> Did-you have seen a lot of topics from non-spammers with URL in subject ? Half of messages posted by spammers has just URL in title, so it's the reason of this change.

Perhaps this is only an issue in **_your_** forum, but it's not an issue in **_everyone's_** forum.  The change, as you've proposed it, would apply to everyone who uses Kunena and this idea should not be universally included as part of the _Kunena_ core product.  It would be "better" (perhaps) to add this "feature" (if people wanted it) as a plugin.

> Nobody is banned, i had added this string but the ban has been removed

If you've added an unnecessary translation string, you should remove the implication that someone would be banned.  Can you imagine:  the website owner creates a topic that contains a URL in the subject and they're banned from using their own forum?  What a bad joke that would be!

No, this idea is unnecessary for **_all_** users of _Kunena_ forums.  From the looks of this there is only one person who has asked for this new feature.  The so-called "spam" problem will not be cured by including additional checks in the _Kunena_ core; additional checks, like these, will cause more problems than they will "cure":  additional lines of code only increase the potential for things to fail, they decrease performance, and they will add to confusion when people start asking questions.

If you want to do this then go ahead.  But don't say that I didn't warn you.

This really isn't a good idea, sorry.

rich20:
- [x] System Messages are correctly
- [ ] URLs within the text are not prevented such as "please visit www.kunena.org". 
  In principle, a space before the url is enough to publish the message.

rich20:
- [x] System Messages are correctly
- [x] URLs within the text are not prevented such as "please visit www.kunena.org". 
  In principle, a space before the url is enough to publish the message.

This feature works only  with addresses such as www.kunena.org. Not, when before is http:// or https://
- [ ] URLs with beginning http:// or https:// are not prevented

rich20:
Now it is reversed
- [x] URLs with beginning http:// or https:// are prevented
- [ ] URLs with beginning www are not prevented

rich20:
Links are prevented
- [x] at New Topic
- [x] at Reply
- [x] at Quick Reply
- [ ] The setting in the backend has no effect - Links are always prevented

src/components/com_kunena/controllers/topic.php
```diff
@@ -482,6 +482,16 @@ public function post()
 			}
 		}
 
+		$url_subject = $this->checkURLInSubject($message->subject);
+
+		if ($url_subject)
```
810: change to `if ($url_subject && $this->config->url_subject_topic)`

src/components/com_kunena/controllers/topic.php
```diff
@@ -705,6 +714,16 @@ public function edit()
...
```
810: same here:  `if ($url_subject && $this->config->url_subject_topic)`

rich20:
Links are prevented
- [x] at New Topic - Success
- [x] at Reply - Success
- [x] at Quick Reply - Success
- [x] Setting in the backend Yes or No - Success

PR #3782, opened 2016-04-09
Remove unneeded column announcement

810:


810:
@xillibit this is not needed right?

xillibit:
Today it's not needed but in the future it can be usefull

sozzled:
xillibit is probably right about the ordering column not needed today but possibly useful in future as far as the table **kunena_announcements** is concerned.  Because _Kunena_ cannot display more than one announcement at the one time, "ordering" has little meaning.

The column was there for the purpose of making a "sticky announcement"—like you can have sticky topics (in the table **kunena_topics**) and, theoretically, "sticky posts" (in the table **kunena_messages**).

Issue #2391, opened 2014-03-21
BBCode Wizard for links does not properly convert square brackets in URLs


c-schmitz:
When entering a link using the bbcode link button in the kunena message editor then URLs with square brackets are not properly converted.

Example Link:
    http://translate.limesurvey.org/projects/limesurvey2/el/default?filters[term]=Only+integer+values

When entered in the wizard you get out:

```
[url=http://translate.limesurvey.org/projects/limesurvey2/el/default?filters[term]=Only+integer+values]Some link[/url]
```

when it should be:

```
[url=http://translate.limesurvey.org/projects/limesurvey2/el/default?filters%5Bterm%5D=Only+integer+values]Some link[/url]
```

Kunena Version 3.0.5

xillibit:
The workaround for now is to add manually quotes like that, tested in K5.2 : 

`[url="http://translate.limesurvey.org/projects/limesurvey2/el/default?filters[term]=Only+integer+values"]Some link[/url]`

PR #3790, opened 2016-04-12
BBCode Wizard for links does not properly convert square brackets in URLs #2391 

810:
Pull Request for Issue #2391 .


rich20:
This bug no longer exists

PR #3812, opened 2016-04-17
Set minimum version to PHP 7.0.3

xillibit:
Pull Request for Issue # .

### DIFF
--- a/src/administrator/components/com_kunena/install/install.script.php
+++ b/src/administrator/components/com_kunena/install/install.script.php
@@ -17,11 +17,11 @@ class Com_KunenaInstallerScript
 {
 	protected $versions = array(
 		'PHP' => array (
-			'7.0' => '7.0.4',
+			'7.0' => '7.0.3',
 			'5.6' => '5.6.8',
 			'5.5' => '5.5.13',
 			'5.4' => '5.4.13',
-			'0' => '5.4.35' // Preferred version
+			'0' => '7.0.3' // Preferred version
 		),
 		'MySQL' => array (
 			'5.1' => '5.1',
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@
 *Kunena* 5.0 requires
 
     Joomla: version 3.5.0 or greater (>= 3.5.0 recommended)
-    PHP: version 5.4.35 or greater (>= 7.0.4 recommended)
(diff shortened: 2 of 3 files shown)


sozzled:
Most people are running with the same PHP minimum version required for Joomla.  Setting the minimum version of PHP to 7.0.3 will discourage people from upgrading to K 5.0.

In other words, requiring people to use PHP 7.0.3 will turn people away from the Kunena project.

Bad idea.

810:
@sozzled Older php 7 versions has security leaks, so we move to  min php to 7.0.3

xillibit:
That's not prevent to use Kunena on php 5.4.x, 5.5.x or 5.6 series, the title is a bit incorrect

src/administrator/components/com_kunena/install/install.script.php
```diff
@@ -17,11 +17,11 @@ class Com_KunenaInstallerScript
 {
 	protected $versions = array(
 		'PHP' => array (
-			'7.0' => '7.0.4',
+			'7.0' => '7.0.3',
 			'5.6' => '5.6.8',
 			'5.5' => '5.5.13',
 			'5.4' => '5.4.13',
-			'0' => '5.4.35' // Preferred version
```
coder4life, line 24: I disagree with this changed. Preferred version is the minimum number that the software should run on. The software will throw warnings everytime you try to install on any 5.6, 5.6, and 5.4 version of the software.

810, line 24: That's correct for me, the users needs some time to update their environments, if we didn't notice them they we be mad. Now we can say, we told you some time to update your environment.

coder4life, line 24: Although that sets a precedent on what the person should be running versus what is actually needed. Just an FYI most shared hosts where a majority of Kunena installs will be running on do not support PHP 7 yet. Many are pushing to 5.5 and 5.6. The preferred version should focus on two things: 1. the minimum version Kunena is absolutely required to run on 2. the End of Life conditions for older PHP versions. I believe on most frameworks PHP 5.5 is the minimum version required.

810, line 24: You only get this notice when you have <5.3.10 & 5.4.0 < 5.4.13 ect they have all low numbers. So you should not see this error, only on bad hosts, then you need to switch immediately

Issue #3815, opened 2016-04-18
Category Subscriptions can't be individually removed in the profile


rich20:
#### Steps to reproduce the issue

Subscribe more than one category
Go in your Profile -> Tab Subscriptions and try unsubscribe a single category
#### Expected result

A single checkbox can be marked
#### Actual result

It marks all check boxes
https://www.kunena.org/forum/kunena-support-reporting-defects/136823-crypsisb3-styling
#### System information (as much as possible)

Joomla version: 3.5.1 an 3.5.2 dev
Kunena version: 4.0.11 dev and 5.0.0 Beta 1
Php version: 5.6 and 7.0.4
Database version: 5.5

PR #3820, opened 2016-04-19
Category Subscriptions can't be individually removed in the profile #3815

810:


rich20:
- [x] Success

You can remove individually categories now but it is missing the checkbox for All. Is it a other bug?

PR #3838, opened 2016-04-21
When save new topic in th logs the mesid is null

xillibit:
Pull Request for Issue # .


rich20:
- [x] Success - it's the same id as in the database now

Issue #3846, opened 2016-04-26
Restore Unapproved Topics not working as expected


xillibit:
#### Steps to reproduce the issue

Delete a topic with the status "Unapproved"
Go to the trashbin and restore the topic.

The expected behavior is that the topic should be restored as "Unapproved"
and it is being restored as Approved. 
#### Expected result

Topics should keep the status unapproved
#### Actual result

Topics has the status approved
#### System information (as much as possible)

Joomla version: 
Kunena version: 4.0.10
Php version:
Database version:
#### Additional comments

See : https://www.kunena.org/forum/kunena-support-reporting-defects/136898-k-4-0-10-restore-unapproved-topics-not-working-as-expected#174538

PR #3856, opened 2016-04-27
Restore Unapproved Topics not working as expected #3846

xillibit:
Pull Request for Issue #3846.


rich20:
- [x] If you move the complete topic to the trash and restore it, pending messages remain pending messages
- [ ] If you delete only a single pending message - after restore the message is approved

rich20:
But I mean, if you delete a single message, you have also checked the message content

xillibit:
For a single message it's useless because the unapproved state is lost when you delete the message

rich20:
- [x] If you move the complete topic to the trash and restore it, pending messages remain pending messages
- [ ] If you delete only single pending messages (see image) - after restore the messages are approved
  bildschirmfoto vom 2016-04-27 23 23 54

Issue #3934, opened 2016-05-11
 Call to a member function get() on string when PHP7 is enabled


Sulpher:
#### Steps to reproduce the issue

I moved site with Kunena 4.0.10 from localhost to hosting. Installed Kunena 5beta4 over 4.0.10. Then  overwrote files at the server with new ones taken from Github branch.
#### Expected result

Forum must work.
#### Actual result

As a result, when PHP 7.0.6 is enabled, forum does not working.

Forum home page shows: 

> Call to a member function get() on string
>  (no Joomla template loaded)

Discussion page shows: 

> 404 Not Found
> You do not have permissions to access this page.
> (Joomla template loaded with Kunena menu) 

If I turning PHP to old 5.4, forum works.
Kunena Latest module is disabled.
#### System information (as much as possible)

Joomla version: 3.5.1
Kunena version: 5 beta 4
Php version: 7.0.6
Database version: 5.5.37-MariaDB-wsrep
#### Additional comments

Could this issue occurs due to server configuration?

810:
There is a change from PHP 7.0.5 to PHP 7.0.6
it has something todo with
public function __get($property)

but I haven't working yet

810:
see: https://bugs.php.net/bug.php?id=72117

mahagr:
To me the error looks like that $category->params is a string, not Registry object. Not sure if it is related to the above change in PHP or not.

Twincarb:
I am having a similar issue
php 7.0.6
The category page loads correctly but unable to click to open a category to view the topics.

Notice: Uninitialized string offset: 0 in /var/www/vhosts/mysite.org/test.mysite.org/libraries/kunena/html/html/kunenaforum.php on line 193

sakicnet:
Please see [https://www.kunena.org/forum/kunena-support-reporting-defects/137045-php-7-0-6-and-kunena-4-0-10-and-5-0b4#175137](https://www.kunena.org/forum/kunena-support-reporting-defects/137045-php-7-0-6-and-kunena-4-0-10-and-5-0b4#175137)

mahagr:
The change will make the number of SQL calls to explode, so I don't recommend doing it in that way.

xillibit:
There is something strange on PHP 7.0.6, in the constructor of the category there is `$registry = new JRegistry();
        if (!empty($this->params))
        {
            $registry->loadString($this->params);
        }
        $this->params = $registry;`

[...]

PR #3942, opened 2016-05-13
Call to a member function get() on string when PHP7 is enabled #3934

810:


mahagr:
This change isn't good. I'll have a look into it...

Issue #3981, opened 2016-06-02
Nobody can see confidential text (also not admins) - but if you click on quote, all users can see it


rich20:
#### System information (as much as possible)

Joomla version: 3.5.1 and 3.6.0-alpha2-dev
Kunena version: 5.0.0-RC-DEV
Php version: 5.6
Database version:

PR #3985, opened 2016-06-03
Nobody can see confidential text (also not admins) - but if you click on quote, all users can see it #3981

810:


810:
@rich20 could you test this one, also check if the confidential is removed on email.

rich20:
- [x] Author, Admins and Moderators can see confidential texts
- [x] E-Mail - No confidential informations
- [ ] With quote button all users can see it

810:
the other issue, is also on K4.0 and other versions

810:
Please create a new ticket for that issue. I'm merging this pr

Issue #4034, opened 2016-06-16
CrypsisB3: poll icon in editor not hidden on default load


betteryouthanme:
#### Steps to reproduce the issue

default installation of Kunena (so you have two categories, Welcome Mat and Suggestion Box)
Now disable polls in Welcome Mat
Create new topic
Category Welcome Mat is load by default (most likey because it's the first in line)
#### Expected result

poll icon is hidden
#### Actual result

poll icon is shown
![16-06-2016 09-09-34](https://cloud.githubusercontent.com/assets/8537297/16108202/2370f45a-33a2-11e6-97c7-c73e3063dd46.jpg)
#### System information (as much as possible)

Joomla version: 3.5.1
Kunena version: 5.0.0-RC2
Php version: 5.6.8 
Database version: 5.6.24
#### Additional comments

CrypsisB2 works fine

xillibit:
I will check that but there is the same js into Crypsis B2 and Crypsis B3

PR #4106, opened 2016-07-07
CrypsisB3: poll icon in editor not hidden on default load #4034

810:


xillibit:
Why do-you remove the load of file pollcheck.js ?

810:
is now on line 51, it was loaded to early.

PR #4116, opened 2016-07-08
Fix check to make difference between image ad file in attachments

xillibit:
Pull Request for Issue # .
#### Testing Instructions

When you do a reply or new topic, add or drag & drop attachments (image or file) and check that the error message which said that the file or image has exceed the limit set in configuration. 

Try by setting in Kunena configuration differents limit for the size for image and file.

### DIFF
--- a/src/libraries/kunena/upload/upload.php
+++ b/src/libraries/kunena/upload/upload.php
@@ -329,7 +329,7 @@ class KunenaUpload
 
 				$size += $bytes;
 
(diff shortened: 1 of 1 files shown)


xillibit:
I have just added testing instructions

@rich20 : can-you test it ?

rich20:
- [x] Images - Attachment **image size** exceed limit allowed by configuration
- [x] Files - Attachment **file size** exceed limit allowed by configuration

Issue #4099, opened 2016-07-06
Prevent to have utf-8 characters in attachment filenames


xillibit:
#### Steps to reproduce the issue

See : https://www.kunena.org/forum/kunena-support-reporting-defects/134454-why-kunena-4-x-allows-utf-8-characters-in-attachment-filenames?limitstart=0
#### System information (as much as possible)

Joomla version: 
Kunena version:
Php version:
Database version:

PR #4120, opened 2016-07-11
Prevent to have utf-8 characters in attachment filenames #4099

xillibit:
Pull Request for Issue #4099  .


810:
Should be a option, he is the only one with that problem, but it is a server setting issue.

see https://www.akeebabackup.com/support/akeeba-backup-3x/20043-chinese-filenames-don-t-backup.html

xillibit:
I think i will a tool to remove all utf8 characters in filename else just an option won't prevent the issue on next extraction of backup

Issue #4117, opened 2016-07-09
Image edit shows only image place holder


rich20:
#### Steps to reproduce the issue
1. Set in configuration "Protect Attachments = No" (default setting)
2. Add a image in a message and submit
3. Edit this message and click on Attachments
#### Expected result

Image preview
#### Actual result

It shows only place holder
#### System information (as much as possible)

Joomla version: 3.5.1 and 3.6.0-rc2
Kunena version: 5.0.0-RC5-DEV
Php version:
Database version:
#### Additional comments

If you set in configuration "Protect Attachments = Yes" then it works.

PR #4132, opened 2016-07-15
Image edit shows only image place holder #4117

xillibit:
Pull Request for Issue #4117  .


810:
Still failing, because missing root.

change it to `JURI::root(true) . '/' . $attach->getUrl();`

xillibit:
When i had tried it has worked like that, i will change it like you said

PR #4135, opened 2016-07-15
Fix rename same filenames

810:


810:
maybe not needed, can't reproduce the issue: https://www.kunena.org/forum/kunena-support-reporting-defects/137773-problems-with-image-attachments

Issue #4138, opened 2016-07-16
Show string, if no maps api key is set.


810:
#### Steps to reproduce the issue

No you see a error on the google map, you always need a key from now. so we need to disable the map, until the key is set.
#### System information (as much as possible)

Joomla version: 
Kunena version:
Php version:
Database version:

PR #4141, opened 2016-07-16
Show string, if no maps api key is set. #4138

xillibit:
Pull Request for Issue #4138 .


810:
change also crypsisb3

xillibit:
Changed too for Crypsis B3

Issue #3744, opened 2016-03-21
Auto embed soundcloud


810:
#### Steps to reproduce the issue

add url
#### Expected result

embed soundcloud
#### Actual result

link
#### System information (as much as possible)

Joomla version: 
Kunena version:
Php version:
Database version:

PR #4142, opened 2016-07-16
Auto embed soundcloud #3744

xillibit:
Pull Request for Issue #3744 .


810:
This is half working.

If you have already done the soundcloud by bbcode, then the bbcode doesn't work.

Because the link is 2x embed, so it will break. We need a check if the url isn't used by the bbcode

xillibit:
Modified to make it working

Issue #3878, opened 2016-05-05
Feature: frontend administration


810:
#### Steps to reproduce the issue

Bring back frontend administration (categories)
#### Expected result

Create/edit/delete categories
#### Actual result

Removed on K1.6
#### System information (as much as possible)

Joomla version: 
Kunena version:
Php version:
Database version:

PR #4156, opened 2016-07-18
Feature: frontend administration #3878

810:


810:
@xillibit could you add the categories on the view, I can't see any category, but on the page load time, I can see its loaded.

Issue #4161, opened 2016-07-19
Ban with delete all post doesn't delete all post (it deletes only the number of Topics Per Page)


rich20:
#### Steps to reproduce the issue

https://www.kunena.org/forum/kunena-support-reporting-defects/137811-ban-with-delete-all-post-doesn-t-delete-all-post
**Backend**
Choose in the configuration (tab frontend) the number of Topics Per Page = 3
**Frontend**
New Ban and select Delete all posts
#### Expected result

All posts are deleted
#### Actual result

Only 3 messages are deleted 
#### System information (as much as possible)

Joomla version: 3.5.1, 3.6.0 and 3.7.0-dev
Kunena version: 4.0.11 and 5.0.0-RC5-DEV
Php version:
Database version:

xillibit:
I will look later

PR #4167, opened 2016-07-20
Ban with delete all post doesn't delete all post (it deletes only the number of Topics Per Page) #4161

810:


rich20:
- [ ] Failed - If you set Topics Per Page = 3 it deletes 3 messages

xillibit:
I'am looking to it if i can found a better fix

Issue #4178, opened 2016-07-22
Email: subscription reply is missing host name


810:
#### Steps to reproduce the issue

https://www.kunena.org/forum/k5-0-bugs/137868-kunena-5-rc5-notifications-problem
#### System information (as much as possible)

Joomla version: 
Kunena version:
Php version:
Database version:

PR #4183, opened 2016-07-24
Email: subscription reply is missing host name #4178

xillibit:
Pull Request for Issue #4178  .


810:
you need to also change it on the system template.

PR #4196, opened 2016-07-26
Fix login users with spaces

810:


src/components/com_kunena/controllers/user.php
```diff
@@ -452,16 +452,20 @@ public function login()
...
 
-		$username  = JFactory::getApplication()->input->post->get('username', '');
-		$password  = JFactory::getApplication()->input->post->get('password', '', 'raw');
-		$remember  = JFactory::getApplication()->input->post->get('remember', false);
-		$secretkey  = JFactory::getApplication()->input->post->get('secretkey', null);
+		$app    = JFactory::getApplication();
+		$input  = $app->input;
+		$method = $input->getMethod();
+
+		$username  = $input->$method->get('username', '', 'USERNAME');
```
xillibit, line 459: You can do directly `$this->app->input` , why you call `$input->$method` ?

810, line 459: im using the code from joomla itself, I just copied the code. see: https://github.com/joomla/joomla-cms/blob/162d10458cd50387b768f6bf5113836359386b34/components/com_users/controllers/user.php#L39

xillibit, line 459: I didn't know that we can do like that

PR #4262, opened 2016-08-09
On update delete directory topic_iconcs to prevent error on files

xillibit:
Pull Request for Issue # .
#### Testing Instructions

extract


810:
I have also done this pr, I think we can better move the dir, if users are custom images, then its lost. So I moved them to a new dir archive

xillibit:
ok, i will delete this one

Issue #4248, opened 2016-08-07
Upgrade from K4 to K5 : Warning Destination folder already exists


rich20:
#### Steps to reproduce the issue

https://www.kunena.org/forum/k5-0-support/138107-warning-destination-folder-already-exists
You can test it:
Downgrade to K4.0.11
then upgrade to K5. Now you see this error. 
![bildschirmfoto vom 2016-08-07 21-15-20](https://cloud.githubusercontent.com/assets/888847/17465140/e807543c-5cef-11e6-8965-7cdbf96358ea.png)

Rename or remove the folder topic_icons in media/kunena and the installation works.
#### System information (as much as possible)

Joomla version: 
Kunena version:
Php version:
Database version:

810:
how to get this error, because we need to have that dir

PR #4263, opened 2016-08-09
Upgrade from K4 to K5 : Warning Destination folder already exists #4248

810:


rich20:
Upgrade works without error but the icon folder remains empty.

Frontend shows this errors:
Warning: simplexml_load_file(): I/O warning : failed to load external entity "../media/kunena/topic_icons/default/topicicons.xml" in libraries/kunena/template/template.php on line 1014

Notice: Trying to get property of non-object in libraries/kunena/template/template.php on line 1031

810:
we need to have the topic_icons dir, closing this one

PR #4320, opened 2016-08-14
Enabled Gravatar Defalt Image feature

demis-palma:
#### Summary of Changes

Enabled Gravatar Defalt Image feature ([further details](https://gravatar.com/site/implement/images/#default-image)).
It's the default image to serve up when there is no Gravatar image associated with the requested email.
The backward compatibility is guaranteed by the default parameter "false".
#### Testing Instructions

Configure plugin Kunena Gravatar and specify your image URL or one of the supported default options: identicon, monsterid, wavatar or retro.
Gravatar API would support 3 additional options (404, mm and blank) but they are not noteworthy to the public.


810:
thnx, will merge it when i tested the pr

Issue #4342, opened 2016-08-17
Blank space when breadcrumb is long


sshcli:
#### Steps to reproduce the issue
1. Create a new section named: Main Forum
2. Create a new category named: Suggestion Box Long Name Category Testing the breadcrumb
3. Create a new sub-category named: Subcategory of suggestion box
4. Create a new topic named: testing breadcrumb with a regular topic title 
#### Expected result

The breadcrumb should fit in 1 single line. 
#### Actual result

When the breadcrumb is long it will take 2 lines and leave a blank space.
![breadcrump](https://cloud.githubusercontent.com/assets/12942731/17736381/84ab3fe4-6455-11e6-9199-59c43f4d67fb.PNG)
#### System information (as much as possible)

Joomla version: 3.6.2
Kunena version: 5.0.1-dev
Php version: 7.0.9
Database version: 5.6.31

sshcli:
This can be solved by just adding an additional field named: Breadcrumb Name
So people can always set a short name for the breadcrumb, even if the category name is very long
<img width="261" alt="bread" src="https://user-images.githubusercontent.com/12942731/37970910-a8ee80ac-31a2-11e8-93d4-7014a6ade5c1.PNG">

PR #4343, opened 2016-08-17
Blank space when breadcrumb is long #4342

810:


810:
@xillibit could you add options, on the template settings, so users can choose max numbers of characters. And add the options on the main.js file

xillibit:
I will do the change to add an option in the evening

sshcli:
Should we close https://github.com/Kunena/Kunena-Forum/issues/4342 too ?

PR #4367, opened 2016-08-20
Final replacement of last calls of remaining JRequest

xillibit:
Pull Request for Issue # .

### DIFF
--- a/src/administrator/components/com_kunena/controllers/categories.php
+++ b/src/administrator/components/com_kunena/controllers/categories.php
@@ -441,11 +441,11 @@ class KunenaAdminControllerCategories extends KunenaController
 			}
 
 			$success = $category->save();
-			$aliases = explode(',', JRequest::getVar('aliases_all'));
+			$aliases = explode(',', $app->input->getString('aliases_all'));
 
 			if ($aliases)
 			{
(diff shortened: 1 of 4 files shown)


src/administrator/components/com_kunena/controllers/categories.php
```diff
@@ -441,11 +441,11 @@ protected function _save()
 			}
 
 			$success = $category->save();
-			$aliases = explode(',', JRequest::getVar('aliases_all'));
+			$aliases = explode(',', $app->input->getString('aliases_all'));
 
 			if ($aliases)
 			{
-				$aliases = array_diff($aliases, JRequest::getVar('aliases', array(), 'post', 'array'));
+				$aliases = array_diff($aliases, $app->input->post->getArray(array('aliases' => '')));
```
810, line 448: with this change all options are going to be removed.

test:
adding some aliases
when you have 4 aliases, then select only the last 3, then click save. Result. only last one is there left

xillibit, line 448: I will re-test with more aliases but in my tests it has the same data the JRequest part

810, line 448: yes, with jrequest its the same bug, so i will merge this one, but we need to look where the bug is

PR #4384, opened 2016-08-22
Fix mb_substr on subject titles

810:


mahagr:
mb isn't installed to all Joomla sites. Joomla has its own JText class for this.

810:
I can't see any substr code their.

mahagr:
Yea, could be. Still you cannot use it. Maybe use preg_replace with `'/.{0,' . $limit . '}/|ui'` instead?

810:
That didn't work, found a other fix

mahagr:
Do not add code to template files. create a function into kunena instead...

810:
moved the files

PR #4448, opened 2016-09-03
Update requirement to Joomla! version to 3.6.2

xillibit:
Pull Request for Issue # .


xillibit:
Updated else it fails

810:
yes this is better

Issue #4465, opened 2016-09-07
Can't search by specific date


sshcli:
#### Steps to reproduce the issue
1. Go to example.com/forum/search
2. Try to search all the posts created on April 29 2016 (Example date, you can use another)
#### Expected result

You should be able to see only the posts created on April 29 2016
#### Actual result

There is no option in the drop down list to choose a specific date
#### System information (as much as possible)

Joomla version: 3.6.2
Kunena version: 5.0.1
Php version: 7.0.10
Database version: 5.6.31

xillibit:
If we want to search at a specific date,  i think we can add a datepicker to choose the date

sshcli:
A datepicker seems to be a good idea. 
Whatever you do will be fine.

PR #4482, opened 2016-09-10
Can't search by specific date #4465

xillibit:
Pull Request for Issue #4465 .


810:
please change also  crypsisb3

810:
At least one keyword should be over three characters long.

I always see this message on the search page

PR #4605, opened 2016-09-27
Improve category icons on index

810:


sshcli:
If possible include this reference:
http://fontawesome.io/cheatsheet/

810:
I have some ideas, but for later.

sshcli:
OK

PR #4615, opened 2016-09-29
Add desc and maxlength to description field (backend)

sshcli:
Pull Request for Issue # .
#### Summary of Changes

Add a description to description field (backend)
Add  maxlength=140 to description field (backend)

desc


810:
Why max 140 ?

sshcli:
That's what you have defined here:
src\components\com_kunena\controller\category\topics\display.php

810:
Thats the value for showing the first 140 characters, if more then the rest will be changed to ...

sshcli:
Ok.
What do you suggest?
Remove maxlength=140 and leave only description?

810:
yes, and remove the (max 140 char) text

Issue #4408, opened 2016-08-24
Backend: fix deleting topics/messages for large sites


810:
#### Steps to reproduce the issue

Delete topics
#### Expected result

deleted
#### Actual result

error 500 or error 524 after some time the query is running
#### System information (as much as possible)

Joomla version: 
Kunena version:
Php version:
Database version:

xillibit:
Error 524 doesn't exist it's only with cloufare when something goes wrong

810:
when removing the recount, i getting this error

Lock wait timeout exceeded; try restarting transaction SQL=DELETE FROM #__kunena_topics WHERE `id` = '136422'

after removing the recount, its works. So we need to move the recount after the function is used.

libraries\kunena\forum\topic\topic.php function delete

cybersholt:
Same problem with:

Joomla version: 3.6
Kunena version: 5.0.1
Php version: 7.0.9
Database version: MySQL 5.7.14

Confirmed that disabling the recount did fix the issue. 

Medium sized forum > 100k users & ~ 550 k posts. 

EDIT: Forgot to mention that the problem exists for the frontend as well.

PR #4616, opened 2016-09-29
Backend: fix deleting topics/messages for large sites #4408

810:


src/libraries/kunena/forum/message/thankyou/helper.php
```diff
@@ -267,11 +267,13 @@ static protected function loadMessages(array $ids)
...
 	/**
-	 * Recount thank yous.
+	 * Recount thank yous is null.
 	 *
 	 * @return boolean|int	Number of rows is successful, false on error.
+	 *
+	 * @since K2.0
 	 */
-	static public function recount()
+	static public function recountthankyou()
```
xillibit: Should be nammed recountThankyou() instead

src/libraries/kunena/user/helper.php
```diff
@@ -530,68 +530,84 @@ public static function recount()
...
-				FROM #__kunena_user_topics
-				GROUP BY user_id
-			ON DUPLICATE KEY UPDATE posts=VALUES(posts)";
+		return $rows;
+	}
+
+	/**
+	 * @return boolean|integer
+	 */
+	public static function recountbanned()
```
xillibit: Should be nammed recountBanned() instead

src/libraries/kunena/user/helper.php
```diff
@@ -530,68 +530,84 @@ public static function recount()
...
 
-		$rows += $db->getAffectedRows();
+		$rows = $db->getAffectedRows();
 
-		// Update banned state
-		// TODO: move out of here, it's slow
+		return $rows;
+	}
+
+	public static function recountpostsnull()
```
xillibit: Should be nammed recountPostsNull() instead

Issue #4643, opened 2016-10-03
Trigger onKunenaGetTopics no more available


obsidev:
#### Steps to reproduce the issue

When we create a menu for Kunena, there is an option "Select View" where we can choose "By Plugin".
That option is not working anymore in Kunena 5.x and the trigger "onKunenaGetTopics" is no more called.
#### Expected result

We are expecting that when the menu is displayed, the trigger is called so we can interact with the listing thanks to a plugin.
#### Actual result

A "my topics" listing is displayed.
#### System information (as much as possible)

Joomla version: 3.6
Kunena version: 5.0.2
Php version: 5.6
Database version: 5.5
#### Additional comments

Patch in progress

PR #4644, opened 2016-10-03
Little improvements in database finder

obsidev:
Pull Request for Issue #4643 .
#### Summary of Changes

Few modifications in the finder object so plugins will be able to modify the content without generating SQL errors.
#### Testing Instructions

When a plugin wants to interact with the finder, it can add some ordering fields which will generate a SQL error during the count.
The patch autorise an empty alias for the "count" function and will also clear the "order" during the "count".


810:
You created patches for the K5.1.0 branch, not for the K5.0.X branch

obsidev:
Due to the modifications (for the 3 PR) I supposed that it would be better for major release than a minor one.

810:
ok, Thnx

PR #4645, opened 2016-10-03
Restore a trigger "onKunenaGetTopics"

obsidev:
Pull Request for Issue #4643 .
#### Summary of Changes

Due to the modification in the Kunena core, some files (models) are not used anymore.
In the new core, the trigger "onKunenaGetTopics" is not more called, so the option "Select View: By Plugin" is not working anymore.

That patch will restore a trigger but will rename it in order to not generate side effect due to the modification in the parameters.
#### Testing Instructions

Create a Kunena menu and select the "By plugin" view mode.
In the header text, put "moderation".

In a plugin, put the trigger:

```
    public function onKunenaGetUserTopics($pluginmode, &$finder, &$order, &$categoryIds, $controller) {
        if($pluginmode != 'moderation')
            return;

        $finder->filterByHold(array(0,1));

        // Force all categories
        $categoryIds = null;

        $finder
            ->select('a.id')
            ->order('last_post_time', 1)
            ->where('a.first_post_userid', '>', '0');
    }
```

_We removed some code for extra columns used by our moderation system_

### DIFF
--- a/src/components/com_kunena/controller/topic/list/user/display.php
+++ b/src/components/com_kunena/controller/topic/list/user/display.php
@@ -107,6 +107,12 @@ class ComponentKunenaControllerTopicListUserDisplay extends ComponentKunenaContr
 				$finder->filterByUser($user, 'subscribed');
 				break;
 
+			case 'plugin':
+				$pluginmode = $this->state->get('list.modetype');
+				$dispatcher = JEventDispatcher::getInstance();
+				$dispatcher->trigger('onKunenaGetUserTopics', array($pluginmode, &$finder, &$order, &$categoryIds, $this));
+				break;
+
 			default :
 				$finder
 					->filterByUser($user, 'involved')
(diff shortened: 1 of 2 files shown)


src/components/com_kunena/controller/topic/list/user/display.php
```diff
@@ -107,15 +107,23 @@ protected function before()
...
 			default :
 				$finder
 					->filterByUser($user, 'involved')
 					->order('favorite', -1, 'ut');
 				break;
 		}
 
-		$categories = KunenaForumCategoryHelper::getCategories($categoryIds, $reverse, $authorise);
-		$finder->filterByCategories($categories);
+		if($categoryIds !== null) {
```
810: please fix codestyle. Add a space after if

obsidev:
I just remember that, in order to work properly, it also require a modification in the "librairies/kunena/forum/topic/finder.php" so the plugin can have access to the "select".

```
    public function select($data) {
        $this->query->select($data);
        return $this;
    }
```

src/components/com_kunena/controller/topic/list/user/display.php
obsidev: Sure

810:
> I just remember that, in order to work properly, it also require a modification in the "librairies/kunena/forum/topic/finder.php" so the plugin can have access to the "select".
>     public function select($data) {
>         $this->query->select($data);
>         return $this;
>     }

Could you also add this on the pr

Issue #4633, opened 2016-10-01
It lacks the delete button in K 5.0.2


rich20:
#### System information (as much as possible)

Joomla version: 
Kunena version:
Php version:
Database version:

810:
the issue is on User Can Delete Own Post option, its seems to be also for mods

rich20:
Yes, confirmed. But also for Admins.

xillibit:
If i understand right, now nobody can see the button delete ?

810:
no, only if you select other then never option

rich20:
> If i understand right, now nobody can see the button delete ?

Yes, you're rght, but there are still more problems.

**User Can Delete Own Post = Always**
- [x] User
- [x] Admin and Moderator

**User Can Delete Own Post = When no Replies**
- [x] User
- [x] Admin and Moderator

**User Can Delete Own Post = Never**
- [x] User cannot delete
- [ ] Admin and Moderator cannot delete

**User Can Delete Own Post = All, except first message of topic**
- [ ] User can nothing delete
- [ ] Admin and Moderator can delete only the first message

PR #4721, opened 2016-10-07
It lacks the delete button in K 5.0.2 #4633

xillibit:
Pull Request for Issue #4633 .


rich20:
- [x] Use Can Delete Own Post = Always
- [x] User Can Delete Own Post = When no Replies
- [x] User Can Delete Own Post = Never
- [ ] User Can Delete Own Post = All, except first message of topic - Failed (User can nothing delete)

xillibit:
> User Can Delete Own Post = All, except first message of topic - Failed (User can nothing delete)

For me it works like it should, i can't delete the first message but the second one i can delete

810:
yes, for me its ok also, maybe rich is a moderator and can see the delete button

rich20:
Yes, it works, maybe it was the cache.

PR #4722, opened 2016-10-07
fix database attachements after remove topics

810:


sshcli:
This will delete the old ones?
Those before 5.0.3?

810:
This will remove the broken attachments on the database. where the topics are deleted.

sshcli:
ok

PR #4723, opened 2016-10-07
Fix error : JUser tryAuthorise does not exists

obsidev:
#### Summary of Changes

There is no "tryAuthorise" function for the JUser class but there is one in the Kunena category.
Regarding the parameters, it looks like it was designed to be the call to the JUser authorise function.
#### Testing Instructions

With an account, edit a Kunena category.
With another account, go in the category listing, in a page where you can see the category under edition.

### DIFF
--- a/src/administrator/components/com_kunena/template/categories/default.php
+++ b/src/administrator/components/com_kunena/template/categories/default.php
@@ -272,7 +272,7 @@ $filterItem = $this->escape($this->state->get('item.id'));
 							<?php
 								echo str_repeat('<span class="gi">&mdash;</span>', $item->level);
 								if ($item->checked_out) {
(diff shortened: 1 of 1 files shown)


810:
its kunena user, not joomla.

See: https://github.com/Kunena/Kunena-Forum/blob/K5.0/src/libraries/kunena/user/user.php#L181

obsidev:
Okay, Kunena User also have tryAuthorise. But the " $this->user " in the view IS a Joomla user object.

It is used like that in other lines in the view. Like:
https://github.com/Kunena/Kunena-Forum/blob/K5.0/src/administrator/components/com_kunena/template/categories/default.php#L205

You just have to do the described test and you will have the same fatal error I had ; error which have this PR as result.

810:
you are right, will merge it.

PR #4730, opened 2016-10-08
Icon updates / .less path fix

betteryouthanme:
Pull Request for Issue # .


src/components/com_kunena/template/crypsisb3/layouts/category/index/default.php
```diff
@@ -38,11 +38,11 @@
 	} ?>
 
 	<div class="kfrontend">
-		<h2 class="btn-toolbar pull-right">
+		<div class="btn-toolbar pull-right">
```
810: please change this to h2 back, we have added h2 because all pages should have 1 h1 and 1 h2 tags

PR #4856, opened 2016-10-23
Attachments upload broken because useless scripts added

xillibit:
Pull Request for Issue # .


810:
we need those files for mp3/video support. after merging new pr from k50, it broke.
Before it was working

PR #4871, opened 2016-10-24
Attachments: Insert all, set all attach as inserted

810:


sshcli:
I still see the problem in  K.5.0.4-dev 2016-10-27

attach

810:
maybe only pictures are inserted, will check the files

810:
for me its working:
twea

sshcli:
Or maybe the problem is nightly builds using old builds

810:
try clean cache and check that the code has changed.

sshcli:
Working now after clean cache

good

PR #4933, opened 2016-11-07
Shorten the word: Reason for editing

sshcli:
Pull Request for Issue # .

#### Summary of Changes
badedit

edit


810:
For english only, then this is ok, but if you have other language, you have still the same issue, we need to fix it right.

PR #5070, opened 2016-11-26
Fixed the wrong URL

demis-palma:
The preview action while editing a post, points to a malformed URL (half SEF, half non-sef), which is not guaranteed to be parsed correctly by Joomla, and alerts some Web Application Firewall since the php file called (/forum/5-support/topic/index.php) does not exists on the filesystem (see below).

Given that the topic URL is http:// site.com/forum/5-support/topic/create.html
and the preview URL is relative index.php?option=com_kunena&view=topic&layout=edit&format=raw
then the actual preview URL will be /forum/**5-support/topic/index.php**?option=com_kunena&view=topic&layout=edit&format=raw which is wrong.

In addition, the value calculated for "kpreview_url" was never used.

### DIFF
--- a/src/components/com_kunena/template/crypsis/assets/js/edit.js
+++ b/src/components/com_kunena/template/crypsis/assets/js/edit.js
@@ -16,7 +16,7 @@ function kPreviewHelper(previewActive) {
 	if (jQuery('#kbbcode-message').val() != null) {
 		jQuery.ajax({
 			type    : 'POST',
-			url     : 'index.php?option=com_kunena&view=topic&layout=edit&format=raw',
+			url     : jQuery('#kpreview_url').val(),
 			async   : false,
 			dataType: 'json',
 			data    : {body: jQuery('#kbbcode-message').val()},
--- a/src/components/com_kunena/template/crypsis/layouts/topic/edit/default.php
+++ b/src/components/com_kunena/template/crypsis/layouts/topic/edit/default.php
@@ -100,7 +100,7 @@ if (KunenaFactory::getTemplate()->params->get('formRecover'))
 		<input type="hidden" name="view" value="topic" />
 		<input id="kurl_topicons_request" type="hidden" value="<?php echo KunenaRoute::_('index.php?option=com_kunena&view=topic&layout=topicicons&format=raw', false); ?>" />
 		<input id="kcategory_poll" type="hidden" name="kcategory_poll" value="<?php echo $this->message->catid; ?>" />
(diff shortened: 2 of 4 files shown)


810:
i had changed this because suffix was breaking the preview. Did you also test it with suffix?

demis-palma:
I'm sorry, I'm not sure to have understood. What suffix? Could you explain further?

810:
The joomla url suffix, like .html

We had a problem with showing the preview, with suffix turned on the joomla global configuration.

demis-palma:
Well, for sure there is a problem with old Joomla 1.6 - 2.5 websites updated to Joomla 3, since they have a statement in their .htaccess which is incompatible with some Kunena URLS, especially the upload:
`RewriteCond %{REQUEST_URI} /component/|(/[^.]*|\.(php|html?|feed|pdf|vcf|raw))$ [NC]`
The statement has been removed in Joomla 3, but old websites often retain their old .htaccess.

That being said, in my opinion, both the code before and after this patch work great regardless of SEF options, because in both cases it points to a non SEF URL.
The problem that this patch wants to solve is that the URL is not completely SEF, still not completely non-SEF. It's a hybrid.
As you can see, it starts with a virtual directory /forum/5-support/topic/... but it ends with a php suffix, instead of the suffix defined in the SEF options.

To conclude, I think that with this patch, the preview should work with every combination of SEF settings, but I haven't tried yet.
Do you want that I make some tests with all the possible SEF settings?

810:
````
Before:
-suffix on and on reply
http://localhost:8081/forum/welcome-mat/1-welcome-to-kunena/reply/index.php?option=com_kunena&view=topic&layout=edit&format=raw
- suffix on and new topic
http://localhost:8081/forum/index.php?option=com_kunena&view=topic&layout=edit&format=raw

After patch:
-suffix on and on reply
http://localhost:8081/forum/welcome-mat/1-welcome-to-kunena/reply/1.html
- suffix on and new topic
http://localhost:8081/forum/newtopic.html

Before:
-suffix off and on reply
http://localhost:8081/forum/welcome-mat/1-welcome-to-kunena/reply/index.php?option=com_kunena&view=topic&layout=edit&format=raw
- suffix off and new topic
http://localhost:8081/forum/index.php?option=com_kunena&view=topic&layout=edit&format=raw

After patch:
-suffix off and on reply
http://localhost:8081/forum/welcome-mat/1-welcome-to-kunena/reply/1
- suffix off and new topic
http://localhost:8081/forum/newtopic
````

810:
@demis-palma thnx, i have also fixed rating and avatar gallery.

PR #5071, opened 2016-11-26
Preview overlap

demis-palma:
Preview element overlaps tabs element. See screenshot.

screenshot3


810:
i had removed it because now you have a big space:
aantekening 1

PR #5106, opened 2016-12-15
Generalized login links

demis-palma:
The option to hide right nav position introduced recently, https://github.com/Kunena/Kunena-Forum/pull/5021 causes that the login links in messages ("Please Log in or Create an account to join the conversation") could point to a non existing login module, so it's necessary to generalise them to use the system login component.


demis-palma:
In other words, if I disable the Kunena login module, all the login links still points to /forum/1-category/2-topic.html#klogin which does not exist.

demis-palma:
Improved to make the code compatible with sites without an existing login menu item.
Tested on both cases, each one in turn, with and without Joomla SEF.

810:
thnx will check it later

Issue #5134, opened 2017-01-04
Meta description is not available in some cases


sshcli:
#### Steps to reproduce the issue
1. Create a new topic and insert the content of following file into the message content: [text.txt](https://github.com/Kunena/Kunena-Forum/files/685699/text.txt)
2. Go to the topic created and click view page source

#### Expected result
Every topic should have a meta description.

#### Actual result
Meta description is not available.

![error](https://cloud.githubusercontent.com/assets/12942731/21656488/b69abebe-d294-11e6-8fba-f28ab94d8aff.PNG)

#### System information (as much as possible)

Joomla version: 3.6.5
Kunena version: 5.0.5
Php version: 7.0.14
Database version: MySQL 5.6.34

sshcli:
@rich20 are you able to reproduce this?

sshcli:
Milestone 5.0.6

PR #5135, opened 2017-01-05
Meta description is not available in some cases #5134

810:


sshcli:
sshcli Test Results: ✔️ Success

PR #5151, opened 2017-01-10
Add missing string

810:


sshcli:
To anyone who knows where is this string being used,  please attach a screenshot. 
By doing this, you will help us (Me and other translators) to have a better idea of the translation word to be used in transifex.
Thanks in advance for your contribution.

810:
its on the new topic modals, like adding polls

sshcli:
Got it, thanks

Issue #5288, opened 2017-04-01
Error message not visible on drag and drop


sshcli:
#### Steps to reproduce the issue
1. Backend->Configuration->uploads->Attachments->Limit Attachments = 1
2. Go to example.com/forum/newtopic 
3. Drag and drop 2 files into message field

#### Expected result
You should see the error without clicking the attachment button

#### Actual result
You must click the attachment button to see the error

#### System information (as much as possible)

Joomla version: 3.6.5
Kunena version: 5.0.7
Php version: 7.1.3
Database version: MySQL 5.6.35

xillibit:
Should i show the error message in modal only when you do drag&drop ?

sshcli:
It could be. I don't see why not
or
Option B = Auto expand the button#kshow_attach_form.btn

or

Option C = show the error in the div.markItUpFooter

sshcli:
in my opinion, option B will solve 3 problems.
Because:
1. Users will be able to see new attachment confirmation without clicking the button.
2. Users will be able to see error messages without clicking the button.
3. Users will be able to see the progress bar (After you add this feature in a later version) without clicking the button. 

Of course, there must be a way to do it with a modal, I just don't know, you are the master.

PR #5297, opened 2017-04-04
Error message not visible on drag and drop #5288

xillibit:
Pull Request for Issue #5288.


sshcli:
I have tested now, and I can't see the difference

810:
remove the fix, and add this to line 252
`$('#kattach_form').show();`

xillibit:
If you do a drag&drop, it show the form, in this case it's work if the file size is above the limit set in confiuration, i will add the change suggested by jelle to make it working in all cases

sshcli:
it works for me now.
but only when moving code to line 252

xillibit:
Try now

sshcli:
**Test results** = Success ✔

PR #5344, opened 2017-05-01
Not possible to upload attachments because check of image/file size is broken

xillibit:
Pull Request for Issue # .

### DIFF
--- a/src/libraries/kunena/upload/upload.php
+++ b/src/libraries/kunena/upload/upload.php
@@ -199,19 +199,18 @@ class KunenaUpload
 			case 'g':
 			case 'gb':
 				$value *= 1024;
-				break;
-
-			// Continue.
+				
+			// Continue , do not put break here
 			case 'm':
 			case 'mb':
 				$value *= 1024;
-				break;
-
-			// Continue.
+				
+			// Continue , do not put break here
 			case 'k':
 			case 'kb':
 				$value *= 1024;
-				break;
+				
+			// Continue, do not put break here
 		}
 
 		return (int) $value;


rich20:
- [x] Success

cyrilreze:
Hello,
This patch is fixing the issue i had on 5.1.0, but is this merged for 5.1.1 ? (i see 5.1.0 Milestone, but this change is not included in this latest release...)

@xillibit thanks for fixing this issue, it saved me time! 👍

810:
this issue has been fix last year already, should be already on K5.1.0

cyrilreze:
@810 ... sorry, it seems library didn't update well on my site... (while all other elements were updated).
I will check this, but maybe an issue on my side...

cyrilreze:
@810 Just tested back with 5.1.0 package (in fact, no issue on install, and the change of this PR is well included on K5.1.0).
The issue is back.
So, the problem does not come from the change of this PR, but the version of upload.php at the time this PR was committed is working.

I didn't have time to check all changes between 5.1.0 version of upload.php, and this one : https://github.com/xillibit/Kunena-forum/blob/683ce6066bae42c99784660d875685279d1875b9/src/libraries/kunena/upload/upload.php
But for me, the latest released version does not work (and block all uploads).

So, not the place here about this issue, but if i have time, i will check if i see where the error comes from in the changes applied between the 2 versions of upload.php file.

Best regards,
Cyril

cyrez:
@810 I took a little time to check my issue, and so open a PR to fix it (dupplicate line in upload.php for M/MB to Bytes conversion) : https://github.com/Kunena/Kunena-Forum/pull/6016

Issue #5346, opened 2017-05-01
Wysiwyg Editor - Insert uploaded image in message is not possible


rich20:
#### Steps to reproduce the issue

There is no way to insert a uploaded image in messages.

#### System information (as much as possible)

Joomla version: 
Kunena version:
Php version:
Database version:

PR #5350, opened 2017-05-02
Wysiwyg Editor - Insert uploaded image in message is not possible #5346

xillibit:
Pull Request for Issue #54346 .


810:
We do still need the old one also.

You can switch editor on : edit template - crypsis - bbcode - Editor

xillibit:
Why you keep the other editor ?

810:
becuase Wysiwyg Editor has some/many issues. So user can always deceit if he want to use the legacy editor. Also for using plugins ect.

xillibit:
I have updated it

Issue #3885, opened 2016-05-05
Feature: ajax upload for avatar


810:
#### Tasks to do
- [x] Allow to add only one image in the form
- [x] Prevent to add files
- [x] Pre-load the avatar on edit
- [ ] Preview the avatar with a circle
- [x] Finish to implement in controller the part which handle the remove of avatar image in ajax
- [x] Enable drag & drop
- [ ] When user click on avatar, it should go directly to edit profile and open the tab to change avatar
- [ ] Let users to choose an image from the gallery

## 
#### Expected result

Ajax upload for avatar
#### Actual result

none
#### System information (as much as possible)

Joomla version: 
Kunena version:
Php version:
Database version:

sshcli:
Before close this issue, please make sure the following problem has been solved: https://github.com/Kunena/Kunena-Forum/issues/4500

810:
@xillibit is the pr ready?

xillibit:
Not yet, i will add on the issue the items that are missing

xillibit:
@810 : are-you able to add the preview as circle on avatar ?

810:
i can't see the avatar now.

could you recheck it on the lasted changes

xillibit:
I will look in the evening

rich20:
- [x] Allow to add only one image in the form
- [x] Prevent to add files
- [x] Pre-load the avatar on edit
- [ ] Preview the avatar with a circle
- [x] Finish to implement in controller the part which handle the remove of avatar image in ajax
- [x] Enable drag & drop
- [x] When user click on avatar, it should go directly to edit profile <strike>and open the tab to change avatar</strike> (The link goes directly to the profile edit but you need a click on the tab to open the avatar options. For me it's ok.)
- [ ] Index page - Click on avatar goes only to the profile
- [x] Other pages - Click goes directly to profile edit
- [ ] Let users to choose an image from the gallery

rich20:
- [x] Allow to add only one image in the form
- [x] Prevent to add files
- [x] Pre-load the avatar on edit
- [ ] Preview the avatar with a circle
- [x] Finish to implement in controller the part which handle the remove of avatar image in ajax
- [x] Enable drag & drop
- [x] When user click on avatar, it should go directly to edit profile <strike>and open the tab to change avatar</strike> (The link goes directly to the profile edit but you need a click on the tab to open the avatar options. For me it's ok.)
- [x] Index page - Click goes directly to profile edit
[...]

PR #5365, opened 2017-05-10
Feature: ajax upload for avatar #3885 (Part 2)

xillibit:
Pull Request for Issue # .

### DIFF
--- a/src/components/com_kunena/controllers/user.php
+++ b/src/components/com_kunena/controllers/user.php
@@ -1010,6 +1010,8 @@ class KunenaControllerUser extends KunenaController
 
 			if (!empty($response->completed))
 			{
+				$this->deleteOldAvatars();
+				
 				// We have it all, lets update the avatar in user table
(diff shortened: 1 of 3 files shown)


810:
error: You can only add one avatar into your profile.

better is to overwrite the avatar

xillibit:
It's normal because you can only add one avatar, you need to click on remove and choose an another one

810:
on the old behavior, the avatar was overwritten.

xillibit:
So i don't keep the remove button and when choose a new avatar it overwritten the old one ?

810:
Keep the delete button, to switch back to default avatar, but upload always allow if req (size) are right.

rich20:
- [x] Avatar upload - Success

Is the avatar galery obsolete now? You can't add images from the avatar galery.

810:
we should keep the gallery

xillibit:
For the gallery i will look in a second time

Issue #5390, opened 2017-05-24
Attachment preview goes wrong in some cases


Sulpher:
#### Steps to reproduce the issue
A message includes attachment. I go to edit this message, then remove existing attachment, upload new one and insert it in message. In preview mode I see no image:
![screen shot 2017-05-24 at 17 31 08](https://cloud.githubusercontent.com/assets/3432048/26415110/08563e2a-40ba-11e7-80dd-fd1f6c4464cc.png)

#### Expected result
Preview of new image.

#### Actual result
No image.

#### System information (as much as possible)

Joomla version: 3.7.2
Kunena version: 5/0.8
Php version: 7.0.19
Database version: 	5.5.52-MariaDB

PR #5402, opened 2017-05-28
Attachment preview goes wrong in some cases #5390

xillibit:
Pull Request for Issue #5390 .

### DIFF
--- a/src/libraries/kunena/attachment/attachment.php
+++ b/src/libraries/kunena/attachment/attachment.php
@@ -237,7 +237,16 @@ class KunenaAttachment extends KunenaDatabaseObject
 			}
 
 			$url = ($thumb ? $fileThumb : $file);
-
+			
+			if (JFactory::getApplication()->isClient('administrator'))
+			{
+				$url = $url;
+			}
+			else
+			{
+				$url = JURI::base() . $url;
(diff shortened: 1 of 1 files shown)


rich20:
The problem is another. If you insert a new image in a message, it will no displayed in the preview. The image can't be displayed, before it is uploaded. But the upload occurs during submit the message.
Therefore you see on edit the previously uploaded images but not the new inserted images.

g4690

xillibit:
I have tried by adding an new attachment image, i haven't checked by adding an image which isn't an attachment

rich20:
When you add a image, the image is only registered in the database. Therefore you have no preview. It will then be uploaded if you click on Submit.  Afterwards the preview works for this image.

xillibit:
In fact it's the reverse, when you submit your message the attachment is saved in database

rich20:
You can test it.
1. Click on reply
2. Add a attachment
3. Click Cancel or Remove Attachment (or do nothing)
4. Check the database or go in backend to "Tools -> Diagnostics"
The attachment is now listed as orphaned

xillibit:
Try now it should works

rich20:
- [x] Success

Issue #5299, opened 2017-04-06
Message "There are no topics to display" is not available in categories


sshcli:
#### Expected result
The message "There are no topics to display" or "COM_KUNENA_VIEW_NO_TOPICS" should be available in the category view. (Crypsis and Crypsisb3)

![yes1](https://cloud.githubusercontent.com/assets/12942731/24764802/87794188-1ac3-11e7-8c40-5245379cc838.PNG)

#### Actual result
The message "There are no topics to display" or "COM_KUNENA_VIEW_NO_TOPICS" is not available in the category view. (Crypsis and Crypsisb3)

![no1](https://cloud.githubusercontent.com/assets/12942731/24764787/7dc72330-1ac3-11e7-8045-64496a39796e.PNG)

#### System information (as much as possible)

Joomla version:  3.6.5
Kunena version: 5.0.7
Php version: 7.1.3
Database version: MySQL 5.6.35

#### Additional comments
The message "There are no topics to display" or "COM_KUNENA_VIEW_NO_TOPICS" may be missing in other views not mentioned above.
For example: Blue eagle use this string on 3 different views
blue_eagle\html\topics\user_embed.php
blue_eagle\html\topics\default_embed.php
blue_eagle\html\category\default.php

sshcli:
I think this is the file to be modified:
/components/com_kunena/template/crypsis/layouts/category/index/default.php

The format used in the other view is this:
<td colspan="4" class="center"><?php echo JText::_('COM_KUNENA_VIEW_NO_TOPICS') ?></td>

sshcli:
Do you plan to do an additional pr for this issue? 
Or should be closed right now?

xillibit:
I will do an another PR to put the text into a box to be more visible

sshcli:
ok, thanks for your response

xillibit:
I didn't found the way to hide this text when there is sub-categories displayed, i will try to look later

PR #5441, opened 2017-06-24
Message "There are no topics to display" is not available in categories

xillibit:
Pull Request for Issue #5299.


sshcli:
This is how it looks now:
now

sshcli:
The problem is solved, but...
Do you think you can move the message above the "New Topic" button ?

sshcli:
I also think, the message should be removed if there are some folders in the category

folders
folders_blue_eagle

xillibit:
> The problem is solved, but...
> Do you think you can move the message above the "New Topic" button ?

The message should be put into box to be more visible

> I also think, the message should be removed if there are some folders in the category

I will try to look to have the same behaviour than blue eagle

PR #5480, opened 2017-07-18
Fatal error Call to undefined method KunenaUser::socialButton()

xillibit:
Pull Request for Issue # .


810:
we need also b/c for 3rt party templates, i think we can better add a new function, so other template builders can still use the old function.

xillibit:
I have modified to keep the old method

sshcli:
Test results = OK✔

Issue #5489, opened 2017-07-26
Unapproved tab not available in crypsis


sshcli:
#### Steps to reproduce the issue
1. UserX creates a new topic in a moderated category
2. ModeratorX try to see all unapproved messages from UserX 

#### Expected result
<img width="534" alt="be" src="https://user-images.githubusercontent.com/12942731/28631900-d0d4579c-71fc-11e7-9a85-9670d3539582.PNG">

#### Actual result
<img width="415" alt="crypsis" src="https://user-images.githubusercontent.com/12942731/28631905-d42eb32e-71fc-11e7-9bb1-52a9b73be468.PNG">

#### System information (as much as possible)

Joomla version: 3.7.4
Kunena version: 5.0.10
Php version: 7.1.7
Database version: MySQL 5.6.35

rich20:
Is strange, I've this tab as moderator or admin only in my own profile. Not in another profile, neither with Crypsis nor with Blue Eagle.

sshcli:
> Is strange, I've this tab as moderator or admin only in my own profile. Not in another profile

Exactly, it should be available for registered users too (You need it If you are a moderator trying to see all unapproved messages from an specific user)

> neither with Crypsis nor with Blue Eagle.

The blue eagle screenshot is from K.4, it means the unapproved tab was available in k.4

810:
Leave a comment

Issue #5381, opened 2017-05-17
Let users see unapproved messages


sshcli:
#### Steps to reproduce the issue
https://www.kunena.org/forum/kunena-feature-requests/139885-k-5-0-6-1-new-feature-request-let-members-see-unapproved-topics-posts

#### Expected result
Users should be able to see unapproved topics / posts, so they can review/edit/delete it before being approved.

#### Actual result
Users are not able to see unapproved topics / posts, so they can't review/edit/delete it once submitted.

#### System information (as much as possible)

Joomla version: 3.7.0
Kunena version: 5.0.8
Php version: 7.1.5
Database version: MySQL 5.6.35

sshcli:
I found this can be also treated as a bug, because is generating 401 access denied errors on google console.

sshcli:
So, maybe the 401 should be fixed in K.5.0
Even when the full enhance will be implemented only in K5.1

PR #5495, opened 2017-07-30
Unapproved tab not available in crypsis #5489

xillibit:
Pull Request for Issue #5489.

### DIFF
--- a/src/components/com_kunena/layout/user/item.php
+++ b/src/components/com_kunena/layout/user/item.php
@@ -35,7 +35,7 @@ class KunenaLayoutUserItem extends KunenaLayout
 		$showSubscriptions = $this->config->allowsubscriptions && $myProfile;
 		$showFavorites     = $this->config->allowfavorites && $myProfile;
 		$showThankYou      = $this->config->showthankyou && $this->me->exists();
-		$showUnapproved    = $myProfile && ($this->me->isAdmin() || KunenaAccess::getInstance()->getModeratorStatus());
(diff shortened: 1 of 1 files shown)


sshcli:
Test results = OK✔

810:
It works, but should it be there at all? because it has nothing todo with the profile.

sshcli:
> It works, but should it be there at all? because it has nothing todo with the profile.

I see your point

**The change is necessary?**
Yes

**Why?**
Because without this pr, there is no option to see all unapproved messages for specific user.

**Can we leave the tab as is?**
The tab It's only visible to moderators, so I think It won't have a negative impact for UX.
Also it was used for a long time in Blue Eagle 4.x without any complaint

**The view should be moved to another place?**
I believe in continuous improvement. It's up to you, it depends of your time to do it.
Enhances are always welcome.

810:
The best way is to show only own unapproved messages, Then we need to allow also topic author to open the topic, then they can also edit it

xillibit:
> The best way is to show only own unapproved messages, Then we need to allow also topic author to open the topic, then they can also edit it

Why you want to allow topic author to open the topic ? Because only moderators and admin can see the unapproved tab

sshcli:
@xillibit I think @810 is talking about this [enhance](https://github.com/Kunena/Kunena-Forum/issues/5381) that will be added in k.5.1

810:
yes, that one i want to include also on the pr.

Issue #5475, opened 2017-07-10
RSS Feed type not working as expected


sshcli:
#### Steps to reproduce the issue
1. Kunena Forum->Configuration->RSS->RSS Feed type = **By Topic**
2. Go to front end and reply a topic in category "Welcome Mat"
3. Go to example.com/welcome-mat?format=feed

#### Expected result
RSS should only show topics 

#### Actual result
RSS is showing posts 

#### System information (as much as possible)

Joomla version: 3.7.3
Kunena version: 5.0.10
Php version: 7.1.6
Database version: MySQL 5.6.35

#### Additional comments
There is no problem with the main rss (example.com/topics/mode-topics?format=feed)
The problem only occurs in rss for specific categories (example.com/welcome-mat?format=feed)

sshcli:
This bug is causing a lot of problems, because there are several services that relies on rss.
I don't know how to fix it.
I will really appreciate your help here.
Thanks in advance

xillibit:
I Will try to look

sshcli:
> I Will try to look

Thanks @xillibit

810:
I can't reproduce this issue, for me its working.

Topics = i see 2 items
Posts = i see 3 items
Recent = i see 2 items

You can't refresh the site, its a different url, so if you use any sef component, then you need to exclude this url.

sshcli:
> if you use any sef component, then you need to exclude this url.

No, I don't use any sef component (Testing environment joomla+kunena only)

> I can't reproduce this issue, for me its working.

What I'm reporting its only the message description, not the quantity of items.
<img width="280" alt="rsswrong" src="https://user-images.githubusercontent.com/12942731/28229805-37930d38-68b2-11e7-8a54-c2cdf9f2f231.PNG">

sshcli:
> You can't refresh the site, its a different url, 

I don't understand what you mean. I'm not refreshing the page.
This url "**example.com/welcome-mat?format=feed**" is found in category index:
<img width="344" alt="secondversion" src="https://user-images.githubusercontent.com/12942731/28230570-8a717e42-68b5-11e7-8757-53170f38ce18.PNG">

xillibit:
I have just tried when open the rss for a category, i can see 3 topics like in the category. In this catgory i have 5 posts

rich20:
I think sshcli is right. There is no difference between **by Topic** and **by Recent Topics**

The descriptions in the backend are:
[...]

PR #5512, opened 2017-08-15
Category rss is now by topic

sshcli:
Pull Request for Issue https://github.com/Kunena/Kunena-Forum/issues/5475

#### Summary of Changes
Before this change the rss for categories where showing the last post. (aka reply)
After this change the rss for categories are showing the first post. (aka topic) 

Note:
RSS for categories doesn't have the option to switch between RSS Feed type (Post,Topic, Recent Topics). 
A new PR should be made in order to add support for this option.


sshcli:
@rich20 @810 @xillibit  Please feel free to improve these changes if needed.

rich20:
Please check this patch again. Opera can't see the news from the subcategories, when is only subscribed the main category. Maybe is included a command, which Opera not know.
The feed shows the first message and the links go to the first message now with all browsers (Chromum, Opera and Firefox). 
Without this patch you see the currently message and also the link goes to the currently message.

sshcli:
> Opera can't see the news from the subcategories, when is only subscribed the main category. 

In my test with Opera, It shows the subcategory always (With and without this patch).
**Without Patch**
withoutpr

**With Patch**
withpr



> The feed shows the first message and the links go to the first message now with all browsers (Chromum, Opera and Firefox).
Without this patch you see the currently message and also the link goes to the currently message.

That's right, the purpose of this pr is to show the topics instead of replies.

rich20:
> In my test with Opera, It shows the subcategory always (With and without this patch).

Maybe the Linux Opera is not identical with Windows Opera. But the ordering of topics is without this patch correctly also with my Opera browser.

> The feed shows the first message and the links go to the first message now with all browsers 
>> That's right, the purpose of this pr is to show the topics instead of replies.

I don't think, users like it, when is subscribed a Feed, and it shows always the same first message.

My opinion:

### By Post : Post - Show the latests posts from the forum

- [x] Global feed : By Post - Shows the latests posts from the forum
- [x] Category Feed : By Post - Shows the latests posts from certain categories

### By Topic - Show the latest topics from the forum (topics only)

- [ ] Global Feed - Should show the last topics without content
- [ ] Category Feed - Should show the last topics fom certain categories without content

**Hint:** The ordering of topics is correct without this patch, it only needs to remove the content
### By Recent topics - Show the latest post from recently updated topics

- [x] Global Feed - Show the latest post from recently updated topics
- [x] Category Feed - Show the last post from recently updated topics from certain categories

sshcli:
> I don't think, users like it, when is subscribed a Feed, and it shows always the same first message.

You are right, the Category Feed should work in the same way as the General Feed.
In this case, We can keep the issue opened, but I will close this PR now. 
Let's wait for a more profesional fix provided by the experts. 
Regards

PR #5524, opened 2017-08-26
Load category template text by ajax

xillibit:
Pull Request for Issue # .


sshcli:
Test results....
The modal to confirm changes is showing up = OK ✔
The "Close" button in modal is working = OK ✔
The "Replace content" button in modal is working = Fail ❌
The "Replace content and keep existing" button in modal is working = Fail ❌
Every text in modal windows has its own language string = Fail ❌

xillibit:
Try now

810:
when i click on "replace content" i get old cat text +new cat text

sshcli:
> when i click on "replace content" i get old cat text +new cat text

I was having the same results as you, but that was was before the last change made by @xillibit 

Now, when I press the "Replace content" or "Replace content and keep existing" button, the text area is cleared and you need to click the category dropdown to insert the template again.

PR #5554, opened 2017-09-13
Add div and class to confidential text for guests

Ruud68:
Currently the text for guests that the message contains confidential information is formatted as regular text in the message. This makes it confusing for quests as they do not understand what this means / why the author typed that.
With this patch the confidential text for guests is placed in a div with class kmsgtext-confidentialguests and by doing that the text can be styled via css to make it look different then the rest of the message.

Pull Request for Issue # .


Ruud68:
see here for example: https://onlinecommunityhub.nl/forum/ochblog/90-canonical-link-issue?start=6#537

Issue #5556, opened 2017-09-14
Configuration -> Show User Status = No | All users get the online status


rich20:
#### Steps to reproduce the issue

Configuration -> Users -> Show User Status = No
With this configuration shows all users with the green label as online

#### System information (as much as possible)

Joomla version: 
Kunena version:
Php version:
Database version:

PR #5582, opened 2017-09-22
Configuration -> Show User Status = No | All users get the online status

xillibit:
#5556

Pull Request for Issue # .


sshcli:
The status has been removed only from the profile view.
It should be removed also from topics and dropdown menu

Issue #5611, opened 2017-10-05
"Report This" button echoed directly from controller


rjhelms:
#### Steps to reproduce the issue

- Within a template override, attempt to modify the position or display of the "Report This" button within Message/Item/Actions.

#### Expected result

- The rendering of the button is done in the template, allowing template developers to customise the display of it.

#### Actual result

- The button is rendered as an echo call within /controller/message/item/actions/display.php, making its positioning and display hard-coded

#### System information (as much as possible)

Joomla version: 3.8.1
Kunena version: 5.0.11
Php version: 7.0.22
Database version: MySQL 5.7.19

#### Additional comments

crypsisb3 includes a call to display the 'report' button using messageButtons within /Message/Item/Actions, but crypsis does not seem to.

810:
the report is not on the message/action but on message/item

https://github.com/Kunena/Kunena-Forum/blob/c49e5e0d8bff4d68ef93cddfb1a7664ba3c8298f/src/components/com_kunena/template/crypsis/layouts/message/item/default.php#L70-L79

rjhelms:
message/item has the modal to complete the report, but the actual "Report this" button is in message/action to my understanding. This seems to be what renders it:

https://github.com/Kunena/Kunena-Forum/blob/c49e5e0d8bff4d68ef93cddfb1a7664ba3c8298f/src/components/com_kunena/controller/message/item/actions/display.php#L308-L311

Other parts of that method add the the button to $this->messageButtons to be handled by the template, but the report button is echoed directly in the controller.

rjhelms:
Appears that this was introduced in #3799

rjhelms:
Appears that this was introduced in #3977.

PR #5612, opened 2017-10-05
Fix report button in Message/Item/Actions

rjhelms:
"Report this" button was echoed directly from controller, rather than passed in to template through messageButtons. This makes it impossible for template developers to modify the display or positioning of the button within the message actions.

Pull Request for Issue #5611.

#### Summary of Changes

Modify message/item/actions controller to include the "report" button in messageButtons, rather than echoing it directly inside the controller.

#### Testing Instructions

- Verify "Report This" button still displays when viewing a thread as a user with adequate permissions.
- Within the template, verify that location and display of the button can be modified.


810:
Thank you for the pr, but i made some changes on it. I'm closing this pr for my own pr.

Issue #5663, opened 2017-12-01
Hide subscribe checkbox on edit


sshcli:
#### Steps to reproduce the issue
1. Edit an existing topic from any user

#### Expected result
The subscribe checkbox should be hidden when editing because it doesn't make sense to be there.
<img width="462" alt="edit2" src="https://user-images.githubusercontent.com/12942731/33505545-16925a48-d6c2-11e7-9aa5-e120b414bb56.PNG">

#### Actual result
The subscribe checkbox is always visible

<img width="468" alt="edit" src="https://user-images.githubusercontent.com/12942731/33505549-1bce9bf2-d6c2-11e7-9f80-b121083a7fc5.PNG">

#### System information (as much as possible)

Joomla version: 3.8.2
Kunena version: 5.0.12
Php version: 7.1.12
Database version: MySQL 5.6.38

sshcli:
Sorry, but I have changed the title and description because I just notice that the subscribe checkbox doesn't work on edit at all, it's not only admin/mod, it doesn't work for anyone.

So, the fix should be hide it for everyone on edit view.

Sorry again for any inconvenience caused

810:
Leave a comment

PR #5664, opened 2017-12-02
Hide subscribe checkbox on edit (Only Mod/Admin) #5663

xillibit:
Pull Request for Issue #5663.


sshcli:
In my test, this PR works ok.
It hides the subscribe checkbox only when admin or global mod is editing someone else topic.

However, I have updated the issue, because I found that the subscribe checkbox doesn't work on edit at all, it's not only admin/mod, it doesn't work for anyone. 

So, the fix should be hide it for everyone on edit view.

Sorry for any inconvenience caused

xillibit:
Subscribe checkbox doesn't appear on edit

PR #5678, opened 2017-12-09
Anonymous fields are displayed even for loggued users

xillibit:
Pull Request for Issue # .


rich20:
Now I'm a bit confused. This option should be available to registered users only. Now can registered users not post anymore as "Anonymous". Or do I understand something wrong?

xillibit:
I think i had misunderstood, i will revert it but not display it for mods/admins

PR #5706, opened 2018-01-14
this changes works better.

diddipoeler:
Pull Request for Issue # .


Sophist-UK:
1. What is the problem this is trying to fix?

2. What does this fix actually do? Please supply before and after screen shots.

3. "This change works better" than what?

4. If this change is merged it will effect everyone. Is this a change that everyone will want or just something that suits you? 

5. Why do you believe that hiding part of the page is better than removing it altogether?

810:
We have stopped support for K5.0 , I will move this into K5.1

PR #5735, opened 2018-02-14
Ranks by CSS description

sshcli:
Pull Request for Issue # .


810:
The option is not only for fontawesome, but also for b2 and b3.

Issue #5729, opened 2018-02-05
Set max length for website name


sshcli:
#### Expected result
<img width="287" alt="goodurl" src="https://user-images.githubusercontent.com/12942731/35825830-614cb30e-0a8d-11e8-9f0a-65e1d291e277.PNG">

#### Actual result
<img width="332" alt="badurl" src="https://user-images.githubusercontent.com/12942731/35825818-5e438566-0a8d-11e8-9a32-161ade818375.PNG">

#### System information (as much as possible)

Joomla version: 3.8.3 and 3.8.4
Kunena version: 5.0.13 and 5.1.0-RC3-DEV
Php version: 7.1.13
Database version: MySQL 5.6.38

#### Additional comments
The title of this issue says: set max length for website name.
But feel free to solve this problem in a more professional way if it possible.

PR #5743, opened 2018-02-17
Set max length for website name #5729

xillibit:
Pull Request for Issue #5729 .


rich20:
- [x] Success - Now allowed maximum 25 character, but it does not reduce already existing names.

Issue #5737, opened 2018-02-16
Possible Cross-Site Request Forgery vulnerability


James-OHara:
#### Steps to reproduce the issue
For a joomla website using kunena-forum navigate to [baseurl]/forum. This is the recent topics view that includes the dropdown form timeselect for selecting the time period to show recent topics from.  

#### Expected result
This form has an anti-csrf token that is included in the form submission and authenticates the token is correct before processing the request.

#### Actual result
This form does not appear to have an anti-csrf token and could be tricking users into accessing a website hosted by the attacker, or clicking a malicious url. More information:
 https://www.acunetix.com/websitesecurity/csrf-attacks/
https://www.owasp.org/index.php/Cross-Site_Request_Forgery
http://www.cgisecurity.com/csrf-faq.html

#### System information (as much as possible)

Joomla version: 3.8.5
Kunena version: 5.0.13
Php version: 7.2
Database version: N/A

810:
Please use the email: security@kunena.org for security issues.

PR #5744, opened 2018-02-17
Add token on timeselect for topics and messages #5737

xillibit:
Pull Request for Issue #5737 .


rich20:
- [ ] It shows no topics at the menu items Recent Topcs, My Topics and No Replies

rich20:
Addendum: If you switch two times the time period, it shows the topics correct. But you must do this every time if you choose another menu item.

810:
@xillibit could you check if this is ok.

xillibit:
Just tested seems to be better

PR #5749, opened 2018-02-19
Update FA Class check circle

sshcli:
Pull Request for Issue # .


810:
please change to far fa-check-circle

Issue #5657, opened 2017-11-28
new menu item: recent posts/topics for unread messages


810:
#### Steps to reproduce the issue
new menu item: recent posts/topics for unread messages

#### System information (as much as possible)

Joomla version: 
Kunena version:
Php version:
Database version:

PR #5786, opened 2018-03-11
new menu item: recent posts/topics for unread messages #5657

xillibit:
Pull Request for Issue #5657.

### DIFF
--- a/src/components/com_kunena/controller/topic/list/unread/display.php
+++ b/src/components/com_kunena/controller/topic/list/unread/display.php
@@ -62,29 +62,11 @@ class ComponentKunenaControllerTopicListUnreadDisplay extends ComponentKunenaCon
 			$controller->redirect();
 		}
 
-		// Get configuration from menu item.
-		$categoryIds = $params->get('topics_categories', array());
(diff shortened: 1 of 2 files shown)


810:
good, only the count for the pagination is wrong also for the count on ( xtopics)

rich20:
The unread topics shows correct, but the numbers of unread topics are wrong. Also if you click "Mark all read" remain the unread numbers the same. No matter which user are logged in. The number of "unread" is always the same.

rich20:
The requested page can't be found.
Call to a member function order() on null

rich20:
The requested page can't be found.
0 Call to a member function start() on null

xillibit:
it's not possible i have removed the line with the order call...

rich20:
The error is different now.

first
Call to a member function order() on null
now
0 Call to a member function **start()** on null

I've first tried with the patch tester. But also created a build with this fix. Both options show the same error.

xillibit:
Try now

rich20:
The error is gone but now show "Unread" the same as the menu item Recent Topics

rich20:
In principle it works correct now. But the count show the number of unread categories.

PR #5868, opened 2018-04-20
Edit modal message

sshcli:
Pull Request for Issue # .

#### Summary of Changes

templatebad

templategood


sshcli:
This PR is an example of how could this message be improved.
I know this is not the most accurate statement for the task in question
if you have an improved version of the modal message, please share it here.
Please keep in my mind that this message will be read by every kind of user, not only administrators.
Thanks in advance for your cooperation

PR #5872, opened 2018-04-27
Remove height: 100% from .kmsg ...

Sophist-UK:
... because it makes every topic / reply the same height as the browser window / screen!!!!!

#### Summary of Changes

Remove `height: 100%` from general.less .kmsg

#### Testing Instructions

Create a topic with several posts. View on high res monitor full screen - lots of white space at the bottom of posts.

Apply fix. Whitespace disappears.

### DIFF
--- a/src/components/com_kunena/template/crypsis/assets/less/general.less
+++ b/src/components/com_kunena/template/crypsis/assets/less/general.less
@@ -398,7 +398,6 @@ form {
 }
 
 .kmsg {
-  height: 100%;
   margin: 20px 0;
   word-wrap: break-word;
   font-family: "Segoe UI", "Segoe UI Emoji", "Segoe UI Symbol", Helvetica , Arial, sans-serif;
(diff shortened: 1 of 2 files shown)


Sophist-UK:
Travis failure is due to missing command sonar-scanner, and not this trivial PR.

810:
i think it was needed for the responsive images

Sophist-UK:
Well if responsive images need some CSS, it needs to be different than this because this has (literally) huge side-effects.

Sophist-UK:
It's weird...

It was my partner who reported it in Chrome on her PC. And I went into developer tools and cleared the checkbox against this CSS and it fixed it. But I couldn't recreate it on the small screen I was using this morning, so couldn't test the fix myself. But I removed it from the live general.less file and she cleared cache and redisplayed and it was fixed. So I created the PR.

But I am now using a bigger screen and have added this same CSS back in using Chrome Developer Tools and it doesn't make the boxes go big.

810:
@Sophist-UK i have made a different approche, showing only to max  959px.

Could you check it with the nightly build.

Sophist-UK:
Sorry - but I don't have a test environment that I can apply it to.

Issue #5873, opened 2018-04-27
Translation strings do not include substitution parameters


Sophist-UK:
#### Steps to reproduce the issue

I wanted to change the topic and reply header strings.

Kunena currently displays them like `{user name} created the topic: {topic}` and `{user name} replied the topic: {topic}`.

I might want to change this to e.g. `"{topic}" created by {user name}` or possibly some languages like German might require a structure something like `by {user name} was "{topic}" created`.

I went to Joomla Language overrides to change the string and found it did not have substitution and was simply `created the topic:`.

This Kunena output is created in `/components/com_kunena/template/crypsis/layouts/message/item/top/default.php` at line 53:
```
echo (!$isReply)
    ? $avatarname . ' ' . JText::_('COM_KUNENA_MESSAGE_CREATED') . ' ' . KunenaForumMessage::getInstance()->getsubstr($this->escape($message->subject), 0, $subjectlengthmessage) 
    : $avatarname . ' ' . JText::_('COM_KUNENA_MESSAGE_REPLIED') . ' ' . KunenaForumMessage::getInstance()->getsubstr($this->escape($message->subject), 0, $subjectlengthmessage);
```

As you can see the avatarname and topic title are in fixed locations rather than being substituted.
Joomla's advice about this is [here](https://docs.joomla.org/Formatted_fields_in_language_translation_strings).

So IMO this should instead be:
```
$title = KunenaForumMessage::getInstance()->getsubstr($this->escape($message->subject), 0, $subjectlengthmessage);
$lngstr = $isReply ? 'COM_KUNENA_MESSAGE_REPLIED' : 'COM_KUNENA_MESSAGE_CREATED';
echo JText::sprintf($lngstr, $avatarname, $title);
```
with the CREATED language string not `created the topic:` but `%s created the topic: %s`.

Then if you wanted to make this `"{topic}" was created by {user name}` you could have a language string `"%2$s" was created by %1$s`.

#### System information (as much as possible)

Joomla version: 3.8.4
Kunena version: 5.1 RC7
Php version: 7.1
Database version: 5.6

#### Additional comments

NOTE: The above is an example - this is probably a Kunena-wide issue, and so presumably for K5.2.

810:
I have added this one crypsis, for some reason it doesn't work for crypsisb3, so i left it to the old behaviour.

PR #5875, opened 2018-04-28
Move more language string concatenations...

Sophist-UK:
... out of PHP and into sprintf of language strings as per #5873.

Note: Rebased to remove non en-GB language string changes. For RTL languages strings (definitely those with parameters, but also those without but with mixed language would benefit) a Unicode RLE character should be inserted at the beginning and a Unicode PDF character should be inserted at the end in order to ensure that the translation is shown the same on LTR and RTL pages regardless of whether the substituted text is LTR or RTL. (Or perhaps this should be included in Joomla JText functionality.)


Sophist-UK:
Please note that I have not been able to test the language strings - as I only have a production site and do not have a multi-lingual test site built.

But an explanation of my thinking behind the RTL is as follows:

Suppose your substitution string is `COM_KUNENA_EDITING_LASTEDIT_ON_BY = "‫עריכה אחרונה: %1$s מאת %2$s.‬"`. Because Github displays RTL unicode correctly, the string is shown with the RTL parts reversed - and characters that are neither LTR or RTL get displayed according to what surrounds them. However the string is actually stored internally more like `‭עריכה אחרונה: %1$s מאת %2$s.‬‬` (which I have displayed here by starting it with a Left-to-Right Override (LRO) character).

If you don't do anything with special Unicode characters, when you substitute the %s with LTR strings you get a string which looks internally like `(RTL1): (LTR1) (RTL2) (LTR2)`. I think that without any special RTL embedded special characters then on an RTL page this will be displayed correctly as `(LTR2) (RTL2) (LTR1) :(RTL1)` but on a LTR page it will be displayed `(RTL1): (LTR1) (RTL2) (LTR2)` which probably makes little sense. `(RTL1): (RTL2) (RTL3) (LTR1)` would be displayed on an LTR page `(RTL3) (RTL2) :(RTL1) (LTR1)` which is a different order - and when the order of words depends on whether the words are in an :TR or RTL character set, that is pretty confusing.

By inserting a special Unicode Right-to-Left-Embedded (RLE) character at the beginning, and a Pop-Directional-Formatting (PDF) character at the end, then it will treat the entire phrase as a single RTL entity and display it consistently.

810:
You should only edit en-gb files, no other languages. They are will be merged by transifex.

Transifex will only merge the en-gb files.

Sophist-UK:
The strings in the other files are identical to existing strings except for the parameters and should not be a problem if they are overwritten. The reason I included them was to avoid users suddenly getting the default English translations where they were previously getting translated versions until Transifex publishes the non-English versions of new strings which might take a couple of transifex cycles (o ne to upload the new English strings and another to download translations.

But if you wish, I can rebase this without the non-English language files. Let me know.

810:
You can better remove those, else there will be maybe a double string will be created, then its harder to debug, also the pr has now conflicts , because i remove the cs_CZ files

Sophist-UK:
Note: I will remove them, however this will lose the special characters I have inserted in the RTL language files.

Can you make Transifex include these as standard?

Sophist-UK:
Rebased to remove non-english translations.

PR #5878, opened 2018-04-29
When you do change vote, the button still appear as vote

xillibit:
Pull Request for Issue # .


rich20:
The error still exist. Change the vote again and you will see it. We have two errors:

- [ ] Change vote again - Kunena Internal Error: but the vote will counted
- [ ] Time Between Consecutive Votes - this time will ignored, you can always to change the vote

Issue #5780, opened 2018-03-07
Change vote second time


810:
#### Steps to reproduce the issue

When you change the vote for the second time you will get a error:

#### Expected result

correct saved

#### Actual result

Kunena Internal Error: Please enable Kunena Debug Mode and report problem in the www.kunena.org forum.

Adding vote to user failed!

The new vote is added but also the old vote still there. First time change is correct, but more, there is a error

#### System information (as much as possible)

Joomla version: 
Kunena version:
Php version:
Database version:

rich20:
Do you mean with "vote" the poll options? I can't reproduce this problem. I've changed it many times, but no error.

810:
set single vote to "no" and 

Time Between Consecutive Votes to 00:00:10

then vote for the first option. 
click on change vote after 10 seconds, select option 2. then you have the error, and both options has 1 vote.

rich20:
ah.. I confirm the Kunena Internal Error

rich20:
The error still exist. Change the vote again and you will see it. <strike>We have two errors:</strike>

- [x] Change vote again
- [ ] Time Between Consecutive Votes - this time will ignored, you can always to change the vote

810:
Should be fixed now, please test

rich20:
Change vote seems fixed now. You see the change button only after the choosen time period. Previously you have no change button. But we have the next bug now. 
If you create a new poll, you get no Vote button. You can to change already exists polls, but in a new poll no voting.

PR #5882, opened 2018-04-30
Change vote second time #5780 (Time Between Consecutive Votes)

xillibit:
Pull Request for Issue #5780


rich20:
These both errors are on all forum pages
Notice: Undefined property: stdClass::$user_id in libraries/kunena/access.php on line 203
Notice: Undefined property: stdClass::$category_id in libraries/kunena/access.php on line 204

This error you get if you click on vote (or change vote)
Notice: A non well formed numeric value encountered in libraries/kunena/forum/topic/topic.php on line 2114

xillibit:
I don't have these errors

810:
I get this one: 
Notice: A non well formed numeric value encountered in C:/Users/Jelle/Documents/gitkraken/Kunena-Forum/src/libraries/kunena/forum/topic/topic.php on line 2114

src/libraries/kunena/forum/topic/poll/poll.php
```diff
@@ -314,7 +314,7 @@ public function getMyTime($user = null)
 			}
 		}
 
-		return (int) $this->mytime[$user->userid];
+		return $this->mytime[$user->userid];
```
810: if you revert this change, the error is gone

810:
i still can switch polls, when set the option to 00:10:10

810:
for some reason i get:
$date_a = new DateTime($poll->getMyTime());  =  public 'date' => string '2018-04-30 20:18:00.000000'
but on my database = 2018-04-30 21:58:48

src/libraries/kunena/forum/topic/poll/poll.php
810, line 317: add the (int) to line 309, then its ok

810:
@rich20 i have fixed it, could you try the last commit

rich20:
Change vote seems fixed now. You see the change button only after the choosen time period. Previously you have no change button. But we have the next bug now.

PR #5937, opened 2018-05-19
Move some js code in backend in separate file

xillibit:
Pull Request for Issue # . 
 
#### Testing Instructions


810:
the src/administrator/recount.js is wrong dir. 
should be
src\administrator\components\com_kunena\template\j3\tools
src\administrator\components\com_kunena\template\j4\tools

Issue #5969, opened 2018-06-03
We need a setting option "User can edit poll if no votes"


rich20:
Currently user can edit their poll when already votes are present.

Joomla version: 3.8.8 and 3.9.0-dev
Kunena version: 5.1.1-DEV
Php version: 7.0.25 and 7.2.1
Database version:

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

PR #5970, opened 2018-06-03
We need a setting option "User can edit poll if no votes" #5969

xillibit:
Pull Request for Issue #5969  . 
 
#### Testing Instructions


rich20:
The function work, but the definition is wrong.

Set you want allow user to edit the poll when someone has voted = No | allow to edit
Set you want allow user to edit the poll when someone has voted = Yes | user can't edit the poll

Issue #6006, opened 2018-06-22
Add option to make email notifications dependent on Joomla user account sendEmail status


pgasiorowski:
**Is your feature request related to a problem? Please describe.**
Yes. Currently Kunena will send email notifications for subscriptions regardless of user's account "Receive System Emails" flag. It would be nice to make this configurable for use case where any email communication must depend on central setting - for instance for GDPR purposes.

**Describe the solution you'd like**
My proposal:
1. Add a Kunena configuration option like "Only email notification if Joomla user account is configured to accepts system emails" or "Treat email notifications as System Emails"
2. Update /libraries/kunena/access.php

Currently the notification list is created in KunenaAccess::getSubscribers()

```php
  $userlist = implode(',', array_keys($userlist));
  $query->where("u.id IN ({$userlist})");

```

I'd update it to:

```php
$userlist = implode(',', array_keys($userlist));
$query->where("u.id IN ({$userlist})");

if(THE_CONFIG_FLAG_IS_ENABLED) {
   $query->where("u.sendEmail = 1");
}

```
this would stop emailing anyone who has opted out from emails.

**Describe alternatives you've considered**
* Make this Kunena account specific - this would make it more difficult to integrate with other components
* Expose emailing to Kunena plugins - to let plugins filter out emails

**Additional context**
Happy to create a PR for this if you consider it a go.

xillibit:
Feel free to make a PR, for me it's ok

pgasiorowski:
Thanks. Do you have a preference for naming, input type and its location?

Would that work for you?
**Location:** Settings > Users > Subscriptions
**Name:** Treat as System Email
**Description:** When enabled, emails will only be sent to users whose Joomla account is enabled to "Receive System Emails"

Preview below:
![screenshot-2018-6-24 kunena forum configuration - my editshare forum - administration](https://user-images.githubusercontent.com/2228236/41817638-4f83b9a4-779f-11e8-836a-fc44940968cc.png)

810:
Looks ok for me

PR #6011, opened 2018-06-27
Add option to use Joomla "Receive System Emails"

pgasiorowski:
Pull Request for Issue #6006
 
#### Summary of Changes 

* New admin config option "Treat as System Email" for subscriptions
* Language strings for en-GB and pl-PL, placeholders for other language packs

#### Testing Instructions

* Update existing kunena/install fresh, subscribe to a thread/category
* Use another account to post in the above thread/category
  * By default or when "Treat as System Email" is set to "No" notifications should be sent to all subscribers
  * When "Treat as System Email" is set to "Yes" notifications will only be sent to subscribers whose Joomla account is configured to "Receive System Emails"


810:
You should only do the en-gb language files, the rest is automatically created by transifex.

pgasiorowski:
Ok. Let me send a fresh PR shortly

PR #6017, opened 2018-06-29
Try new function tobytes

810:


810:
@JoomliC please try this pr

cyrez:
Thanks @810 

Look better solution, even if i don't know which other issues remaining... (EDIT : i see, "memory_limit"...)

I've tested your PR, and it works! 👍

PR #6038, opened 2018-07-12
Bug: Fix empty action dropdown

pgasiorowski:
**Describe the bug**
Empty dropdown appears for unprivileged users.

**To Reproduce**
1. Use regular/unprivileged account
2. Go to list of topics
3. "Select Action" dropdown is empty and has no use

**Expected behavior**
The dropdown should be hidden.

**Actual result**
It has only "- Chose action -" option

**Screenshots**
chooseaction

**System information (please complete the following information)**

Joomla version: 3.8
Kunena version: 5.1.1
Php version: 7

### DIFF
--- a/src/components/com_kunena/models/category.php
+++ b/src/components/com_kunena/models/category.php
@@ -460,7 +460,7 @@ class KunenaModelCategory extends KunenaAdminModelCategories
 			$actionDropdown[] = HTMLHelper::_('select.option', 'restore', JText::_('COM_KUNENA_BUTTON_UNDELETE_LONG'));
 		}
 
-		if ($actionDropdown == 1)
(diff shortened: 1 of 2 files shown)


810:
does the count also work on Php 5.x  and 7.x .

pgasiorowski:
Yes, count() works on 5 and 7, in fact it was added long tim before 5.
I've done an extra test on 5.6.34 to double-check.

FWIW this could also be fixed in the template but I think it's good to fix it at the source as other templates have same issue (eg Kvivid)

rich20:
- [x] Yes, it works as it should.

Tested with php 5.5.38 and php 7.2.1 as well Joomla 3.8.11-dev and Kunena 5.1.2-DEV

Issue #6058, opened 2018-07-24
Add reviewCount beside stars


sshcli:
**Steps to reproduce the behavior:**
1. Configuration->Allow Rating = Yes 
2. CategoryX->Allow Rating = Yes 
3. Go to any topic in CategoryX

**Expected behavior**
Use the "reviewCount" value from "aggregateRating" and make it visible to all users (The same value that is used for json-ld).
This will help users to know how many reviews does have the topic.

![rvgood](https://user-images.githubusercontent.com/12942731/43165343-abb81aa6-8f61-11e8-9d51-ea891f9cc221.PNG)

**Actual result**
![rvbad](https://user-images.githubusercontent.com/12942731/43165350-b056d6e2-8f61-11e8-865d-a48b998f3a1c.PNG)

**System information (please complete the following information)**
Joomla version: 3.8.10
Kunena version: 5.1.3-DEV
Php version: 7.2.7
Database version: MySQL 5.6.39

PR #6062, opened 2018-07-27
Add reviewCount beside stars #6058

xillibit:
Pull Request for Issue #6058  . 
 
#### Testing Instructions


sshcli:
Test results = ✔️Pass

Basically, the function works ok.

Two minor observations to consider:
1. The number needs additional style (Maybe in a second PR)
it could be something similar to Google reviews:
format

2. Banned users are able to rate a topic, even when they get the error message saying they are not allowed (Without this fix banned users can rate the topic without any warning)
Maybe this is a different and unrelated bug. I will create an issue for it after merge this PR.

PR #6093, opened 2018-08-20
fix for generated kunena-custom.css pointing to wrong location

Ruud68:
custom.less issue, after update to 5.1.3 non-existing file (kunena-custom.css location) is loaded...

Pull Request for Issue # . 
 
#### Testing Instructions


Ruud68:
this is only crypsis and not crypsisb3 as I cannot test this change for b3 :)

Issue #6124, opened 2018-10-09
Add a button to subscribe/unsubcribe users to categories/topics


xillibit:
Add a button in backend to subscribe or unsubcribe users to some categories/topics

xillibit:
For categories it's has been done in https://github.com/Kunena/Kunena-Forum/pull/6163

sshcli:
When you click the button  `Add subscriptions to categories` it opens  the modal for `Assign Moderators`

PR #6168, opened 2018-11-26
Add a button to subscribe users to categories/topics #6124

xillibit:
Pull Request for Issue # . 
 
#### Testing Instructions


rich20:
Still a small mistake exists (see image). A global moderator is not a category, nevertheless you get the message: "The subscriptions to categories for users has been added"
bildschirmfoto von 2018-11-26 18-51-04

sozzled:
Why is this feature being considered?  I always thought that category subscriptions was an opt-in feature.  It looks like this feature will allow anyone to subscribe anyone else to categories whether or not the subscribed users actuall _**want**_ to be subscribed.  Is that the intention?

This also means that people who use this feature will need to modify their sites' terms of use policy to inform users that they may receive email notifications about new posts on the forum whether or not they want to be notified.  I think there are more problems here than what the feature is intended to solve.

sozzled:
Let me ask the question another way.  I haven't seen any recent public discussion about the need for this new feature; the discussion died out years ago.  Who is asking for it now?  What is the public's support for it now?  Where can we find some information relating to the **_current_** pressure to implement this feature?

PR #6173, opened 2018-12-02
set forum messages in own div container

Ruud68:
In order for Infinite scrolling to work correct on items view, all items should be listed in its own div container oppesed to as it is now (in container with all other elements, like search, etc.)


Ruud68:
When this one is okay, I can do the PR for K6.0 as well :)

Issue #6112, opened 2018-09-12
Backend: no action when I clicking on Moderator icon


Sulpher:
**Describe the bug**
I go to Site Backend > Kunena > Users and click on Moderator icon. It takes no action.
![screen shot 2018-09-12 at 09 44 47](https://user-images.githubusercontent.com/3432048/45406966-c0adad00-b670-11e8-8281-0058887bbd2a.png)

 

Joomla version: 3.8.12
Kunena version: 5.1.3

810:
We have only show if moderator, it should not set a moderator because of the permissions.

810:
maybe I can add only unmoderated to it.

PR #6221, opened 2019-02-26
Backend: no action when I clicking on Moderator icon #6112

xillibit:
Pull Request for Issue #6112 . 
 
#### Testing Instructions


810:
Maybe a extra check.

example now:
User with permission on 1-3 categories, when you click on the button, he will get global permissions. I think for the normal mod should be unmod on first click, then on the second click global.

Issue #6230, opened 2019-02-27
RSS Feed | Ex- and Include categories affects the entire global feed


rich20:
**Describe the bug**
https://www.kunena.org/forum/general-questions-and-how-tos/157288-global-rss-feed-not-working
I confirm this bug

**To Reproduce**
Exclude or include in the RSS setting a category

**Actual result**
- [ ] Exclude a category - it exclude all categories in the global feed
- [ ] Include a category - shows nevertheless all categories in the global feed

xillibit:
With excluding on category, the RSS feed isn't empty and there is right the topics of other category : 

![config_rss](https://user-images.githubusercontent.com/876395/53652366-0541fa80-3c49-11e9-94da-cf49edb43676.jpg)

What i'am missing ?

rich20:
I'm sorry but I see my mistake now. I was just as irritated as the user in the forum. I saw in the configuration report of this user the excluded categories as text (such as welcome-mat) and therfore also it tried with the name of the categories.
Mayby we should adjust the description in the backend. It's not obvious that you must to use the ID.


>Specify a list of categories (separated by comma) to be excluded in all feeds. 
Leave empty to ignore feature.


>Specify a list of categories (separated by comma) to be included exclusively in global feeds. 
Please note: This option is not real useful, because global feeds will then ONLY present data from these categories specified.
Leave empty to include all categories in global feeds.
Note: categories will be ignored, if feed is on a category level.

PR #6231, opened 2019-03-01
RSS Feed | Check that value given in exclude categories are right int #6230

xillibit:
Pull Request for Issue #6230. 
 
#### Testing Instructions


rich20:
@xillibit 
Is this pull request still up to date? You made this pull request after Jelle closed the issue. 
With this fix  it excludes all categories in the global feed if you enter a category ID.

xillibit:
Yes, it's up to date. It was designed to check if the values entered by the user are right and not text. I will check it

xillibit:
I close it for now, i need to find an another way

PR #6246, opened 2019-03-08
Allow one message per page

sshcli:
#### Summary of Changes 
bad

good


xillibit:
What this change does ?

sshcli:
It let you choose a minimum of one message per page.
Before this change, the minimum messages per pages is two.

Please, See the pictures I have added in description

PR #6247, opened 2019-03-08
Avoid duplicate meta description on replies

sshcli:
#### Summary of Changes 
This small change will prevent duplicate meta description only on replies pages.
This is in cases where google and other search engines shorten description before reach 140 characters.


810:
you can better change the 140 to 130, because now it has no changed behavior except reversed

sshcli:
Ok, Done.

PR #6280, opened 2019-04-10
Email header image is too big

sshcli:
#### Summary of Changes 
Reduce image size in notification and report emails from 560px to 128px
Also remove some unnecessary blank spaces


sshcli:
I have tested this in my laptop, and it works fine.
It will be nice if someone else could test it on different devices (Smartphone, Tablet, Desktop, etc.) before merge.

810:
Why 128, should be a option on the settings.

sshcli:
> Why 128, should be a option on the settings.

I agree, there should be an option
I will create an issue for it and then close this PR

Issue #6306, opened 2019-05-15
Facebook only accepts 52 characters


sshcli:
**To Reproduce**
<img width="308" alt="facebook" src="https://user-images.githubusercontent.com/12942731/57747494-c4804c00-76a3-11e9-82a7-606c65f4d8ec.PNG">

**Expected behavior**
Facebook should accept long profile names

**Actual result**
Facebook only accepts 52 characters. 
Users with long profile names are getting 404 pages

**System information (please complete the following information)**
Joomla version: 3.9.6 and 4.0.0-alpha8
Kunena version: 5.1.12.1 and Kunena 6.0.0-ALPHA3-DEV
Php version: 7.3.5
Database version: MySQL 5.7.26

sshcli:
This is the error In Kunena 6.0.0-ALPHA3-DEV + Joomla! 4.0.0-alpha8:

<img width="593" alt="error" src="https://user-images.githubusercontent.com/12942731/57873518-77ec5c00-77dc-11e9-9b2e-699ff684f74c.PNG">

sshcli:
Problem Persist / Same error

sshcli:
Please Reopen this issue

810:
You need new install or update from K5.x

sshcli:
ok, understood.
I was testing with 6.0.0-ALPHA3-DEV-GIT

xillibit:
When you update from git Joomla! it shows a warning to make a database update, maybe there is a way to do the same in Kunena...

PR #6314, opened 2019-05-17
Facebook only accepts 52 characters #6306

xillibit:
Pull Request for Issue #6306. 
 
#### Testing Instructions


810:
Maybe we should check the other socials also

Issue #6287, opened 2019-04-24
The word URL is hard coded


sshcli:
**To Reproduce**
Take a look to the word "URL" found in notification emails and report emails.

**Expected behavior**
All words in the forum should have a language string.
We should always avoid the use of hard coded words when possible

**Actual result**
The word "URL" is hard coded

**System information (please complete the following information)**
Joomla version: 3.9.5
Kunena version: 5.1.12.1
Php version: 7.3.4
Database version: MySQL 5.7.25

sshcli:
Please assign this one to me, I can do a PR later

PR #6319, opened 2019-05-17
The word URL is hard coded

sshcli:
Pull Request for Issue https://github.com/Kunena/Kunena-Forum/issues/6287


sshcli:
I haven't test this one, because I'm having problems with my mail server right now

Issue #6339, opened 2019-05-21
Database error on User Synchronization


rich20:
There are two different errors on User Synchronization

- [ ] **Mark Remove user profiles from deleted users**
You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'AS `a` LEFT JOIN `rich_users` AS `b`ON a.userid=b.id WHERE b.username IS NULL' at line 2

- [ ] **Mark Delete lifetime ban users**
You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'AS `a` LEFT JOIN `rich_users` AS `b`ON a.userid=b.id WHERE banned='1000-01-01 00' at line 2

Joomla version: 4.0.0-alpha10-dev
Kunena version: 6.0.0-ALPHA3-DEV
Php version: 7.3.3 and 7.3.5
Database version: 5.7.25

PR #6390, opened 2019-05-30
Database error on User Synchronization #6339

xillibit:
Pull Request for Issue #6339. 
 
#### Testing Instructions

### DIFF
--- a/src/administrator/components/com_kunena/controllers/tools.php
+++ b/src/administrator/components/com_kunena/controllers/tools.php
@@ -287,12 +287,8 @@ class KunenaAdminControllerTools extends KunenaController
 
 		if ($userdellife)
 		{
-			$query = $db->getQuery(true)
-				->delete('a')
-				->from($db->quoteName('#__kunena_users', 'a'))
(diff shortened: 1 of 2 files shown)


rich20:
> There are two different errors on User Synchronization

The second error is solved, but the first error is still there

- [ ] **Mark Remove user profiles from deleted users**
You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'AS `a` LEFT JOIN `rich_users` AS `b`ON a.userid=b.id WHERE b.username IS NULL' at line 2

- [x] **Mark Delete lifetime ban users**
You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'AS `a` LEFT JOIN `rich_users` AS `b`ON a.userid=b.id WHERE banned='1000-01-01 00' at line 2

xillibit:
Try now

rich20:
Now we have an error on the others option again
- [x] **Mark Remove user profiles from deleted users**
- [ ] **Mark Delete lifetime ban users**  | Error - Query was empty

xillibit:
Try now

rich20:
- [x] Success

Issue #6596, opened 2019-11-19
Ban users in the frontend is not possible


rich20:
**Describe the bug**
Ban users in the frontend is not possible, but it works in the backend.

**To Reproduce**
- [ ] If you want ban from the forum you get the message:
**User ban successfully removed.**
- [ ] If you want ban from the site you get the message:
**User account successfully enabled.**

**System information (please complete the following information)**
Joomla version: ‎4.0.0-beta1-dev
Kunena version: 6.0.0-ALPHA4-DEV
Php version: 7.3.6
Database version: 5.7.26

xillibit:
This is just a fix related to the issue but it not fix it

xillibit:
> If you ban an user only from the Forum (not from the site) and you set a time where the ban ended, then the user should not have write access before the ban ended. But the user remain a normal user without ban and can nevertheless write posts.

The user banned should be logged out when the mod/admin do the ban

PR #6601, opened 2019-11-24
Ban users in the frontend is not possible #6596

xillibit:
Pull Request for Issue #6596 . 
 
#### Testing Instructions

### DIFF
--- a/src/libraries/kunena/user/ban.php
+++ b/src/libraries/kunena/user/ban.php
@@ -642,7 +642,7 @@ class KunenaUserBan extends CMSObject
 			return;
 		}
 
-		if (!$expiration || $expiration == '9999-12-31 23:59:59')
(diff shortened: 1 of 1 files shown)


rich20:
-  [x] Success

rich20:
hm.. I see now, ban from the forum with end time does not work. Ban from the page with end time works but can see it only tomorrow, whether the ban will automatically removed.

xillibit:
> hm.. I see now, ban from the forum with end time does not work. 

Why do-you mean by it doesn't work ?

rich20:
If you ban an user only from the Forum (not from the site) and you set a time where the ban ended, then the user should not have write access before the ban ended. But the user remain a normal user without ban and can nevertheless write posts.

PR #6605, opened 2019-11-27
Add unique indices to category template id tags

iantrudell:
Using the same id="kpost-category" ids repeatedly is not allowed, so this adds a unique index to the end of the ids

#### Summary of Changes 

Echo the category ID at the end of each id="x" string to ensure each ID tag is unique.


810:
Thank you for the pr, please also do crypsisb3 template. 

Ps you should also use joomla codestyle. (spaces around the dots and =>)

iantrudell:
> Thank you for the pr, please also do crypsisb3 template.
> 
> Ps you should also use joomla codestyle. (spaces around the dots and =>)

Amended to use the Joomal code style and added the change to crypsisb3

PR #6608, opened 2019-11-30
Add unique classes to topic actions toolbars

iantrudell:
The two topic-actions-toolbars shared the same ID, which was not allowed. I've changed the ID to a class and added unique ID's to their parent div's to allow targeted styling and prevent the same ID being used twice on the same page. Should not effect existing styles, but will enable better style control for template creators.
 
#### Summary of Changes 

Remove duplicate ID tags and add unique ones on their parent divs.


810:
for now, i don't want to break custom css. 
Adding id's is ok, but not removing, that one we can do only with big releases.
But we will have new templates.

xillibit:
We aren't accepting PR against 5.1 branch, feel free to re-do a new PR against 5.2 branch

Issue #6619, opened 2019-12-09
Unknown fetch type '8' 


810:
**Describe the bug**
Unknown fetch type '8'

JROOT/libraries/kunena/forum/category/helper.php:101

**To Reproduce**
K6.0

See:
https://github.com/joomla/joomla-cms/issues/24857

xillibit:
Where do-you have this issue now ?

810:
Do git pull, then composer install. The Joomla/database Will be updates. Then you Will have this issue

xillibit:
I have updated Joomla! but not composer, i will try it

PR #6621, opened 2019-12-09
Unknown fetch type '8' loadObjectList() #6619

xillibit:
Pull Request for Issue #6619. 
 
#### Testing Instructions


810:
Do we need also need to change line 57?

Issue #6674, opened 2019-12-28
Syntax error when delete permanently


rich20:
**To Reproduce**
Go in the frontend to Recent Topics
Mark a topic
Choose on bottom the action "Permanent delete selected" and click on Go

**Actual result**
The topic will be deleted but you get also this error:
You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'AS `a` LEFT JOIN *_kunena_messages` AS `m` ON a.mesid=m.id WHERE m.id IS NUL' at line 2

**System information (please complete the following information)**
Joomla version: ‎4.0.0-beta1-dev
Kunena version: 6.0.0-ALPHA5-DEV
Php version: 7.3.11
Database version: 5.7.28

PR #6676, opened 2019-12-28
Syntax error when delete permanently #6674

xillibit:
Pull Request for Issue # . 
 
#### Testing Instructions


rich20:
- [x] Success

810:
why is the query being executed there is no $attachments.

So i think the issue is somewhere else

xillibit:
There is already a check before and if there ar no attachments this part of code whith this query isn't executed

PR #6709, opened 2020-01-05
Remove call of HTMLHelper::_('behavior.tooltip') which doesn't exist

xillibit:
Pull Request for Issue # . 
 
#### Testing Instructions


810:
is libraries/cms/html/bootstrap.php

Issue #5716, opened 2018-01-28
Attachment reference is not being deleted


sshcli:
#### Steps to reproduce the issue
1. Create a topic and attach a file
2. Click the "insert" button, to insert the attachment reference in the text area
3. Go to attachment manager in frontend and delete the file from there

#### Expected result
The attachment reference inserted in the text area should be also deleted automatically.
Just like it does when you delete the file while editing the topic.

#### Actual result
The attachment reference stays intact, showing the error message" Attachment not found"

#### System information (as much as possible)

Joomla version: 3.8.3
Kunena version:  5.1.0-RC1 and 5.0.13
Php version: 7.1.13
Database version: MySQL 5.6.38

xillibit:
I'am not sure if it will possible to do that, to avoid the message "Attachment not found" we need to edit the content of message and removing the right reference

sshcli:
Ok,
I have created this issue because I saw it was possible to remove it automatically when you delete the attachment while editing a topic.

But that's ok if this can't be done.
Please feel free to close this issue if this definitively not gonna happen

xillibit:
> Kunena version: 3.8.3

That's an old Kunena version 😄

sshcli:
> That's an old Kunena version 😄

Edited: Thanks!

rich20:
**Editor**
- [x] Attachments which have been inserted one by one can be removed one by one
- [ ] The button "Remove all" removes nothing
- [ ] Attachments which have been inserted with the button "Insert all" all will be removed if you want only remove a single attachment

**Frontend attachment manager**
- [x] Images which have been inserted one by one will be removed correctly when text is in the message too
- [ ] Images which have been inserted one by one will be removed but when no text is in the message, one of this code remains
- [ ] Attachments which have been inserted with the button "Insert all" will be removed but the code remains in the message

**Backend file manager**
- [ ] The code always remains in the message

rich20:
**Editor**
- [x] Attachments which have been inserted one by one can be removed one by one
- [ ] The button "Remove all" removes the attachments only from the attachment form but not from the text field. After submitting the attachments are present in the message.
- [ ] Attachments which have been inserted with the button "Insert all" all will be removed from the text field if you want only remove a single attachment. You must to insert again the remained attachments (see image).  You see, the text field is empty, but the attachment form shows "In Message"
[...]

PR #6715, opened 2020-01-08
[K5.1] Attachment reference and content deleted on remove/all #5716

xillibit:
Pull Request for Issue #5716 . 
 
#### Testing Instructions


rich20:
If you insert images with the button "insert all" in messages and you remove one of this image with the button "Remove file", you get in the message field the word **false**. When you save then remains only this word in the message and from the rest of the images remain only thumbnails.

xillibit:
@rich20 : can-you try now ?

rich20:
Delete via backend or frontend file manager is for me ok now but only for attachments which have been inserted one by one. 

I don't understand why there is a difference on delete if you insert the attachments one by one or have been added with the button "Insert all".
If you use the button "Insert all" remain all codes, whether indifferent text is in the message or not.

## These issues exists only with attachments which are inserted with the button Insert All
**Edit message**

- [ ] Attachments which have been inserted with the button "Insert all" all will be removed from the text field if you want only remove a single attachment. It adds the word "**false**" in the text field instead.

**Frontend attachment manager**
- [ ] Attachments which have been inserted with the button "Insert all" will be removed but all codes remain in the message

**Backend file manager**
- [ ] Attachments which have been inserted with the button "Insert all" will be removed but all codes remain in the message

xillibit:
> I don't understand why there is a difference on delete if you insert the attachments one by one or have been added with the button "Insert all".

Because it's two diffrents functions :remove with insert all and remove all files

rich20:
There is also the option "Remove attachment from the list". But this option is only available, when the attachments would have been inserted one by one.
You see it on the image, I've added the first attachment with the button Insert All and here lacks this option.
Bildschirmfoto von 2020-04-11 19-59-32

xillibit:
> There is also the option "Remove attachment from the list". But this option is only available, when the attachments would have been inserted one by one.
> You see it on the image, I've added the first attachment with the button Insert All and here lacks this option.

That's is something that @jelle has added but i don't understand exactly the purpose

810:
when the image is inline, then the image isn't showed on the attachment under the message.
Before it showed in the message and under the message on the attachment list.

810:
Please add insert all, on edit message.

Also when insert 3 from 6 images, all images are showed on the thumbnail bellow the message.

xillibit:
> Please add insert all, on edit message.
> 
> Also when insert 3 from 6 images, all images are showed on the thumbnail bellow the message.

I will look at it

810:
ok, thats looking better, next issue: When insert all, skip files that already in the message.

rich20:
The button "Remove All" removes itself if you click on it. It's indifference with which button the attachments have been inserted.
Reproduce it:
1. Add on reply one or more attachments in the message
2. Click on Remove All (it removes only the Remove All button the attachments remain)

xillibit:
> ok, thats looking better, next issue: When insert all, skip files that already in the message.



> The button "Remove All" removes itself if you click on it. It's indifference with which button the attachments have been inserted.
> Reproduce it:
> 
>     1. Add on reply one or more attachments in the message
> 
>     2. Click on Remove All (it removes only the Remove All button the attachments remain)

I will fix that in another PR

Issue #5095, opened 2016-12-12
Custom images for Categories not working properly


LeonelF:
#### Steps to reproduce the issue
When you follow the instructions to replace the Icons of categories by editing the custom.less file (for crypsis template) it doesn't work properly.

The reason for this is because, even if we select Images - Legacy, it uses a <i> tag or a <span> tag with the icon class we set, but that is incorrect, for "normal" images it should be a <div> and it shouldn't load any icon* related class

How the html looks like (doesnt work, blank space and no image):
```
<td class="span1 center hidden-phone">
<a href="/mywebsite/index.php/kunena/mycategory"><span class="mycategory-icon icon-big " alt="Mensagens Novas" align="center"></span></a>
</td>
```

How it should look like (works, image displays as expected):
```
<td class="span1 center hidden-phone">
<a href="/mywebsite/index.php/kunena/mycategory"><div class="mycategory-icon" alt="Mensagens Novas" align="center"></div></a>
</td>
```

#### Expected result
To show our custom image for the category

#### Actual result
Blank space

#### System information (as much as possible)

Joomla version: 3.6.4.
Kunena version: 5.0.4.1
Php version: 7.0.14-1
Database version: 5.7.16

#### Additional comments
I think that it's very easy to resolve this problem that affects so many users just replacing the span or i tag for a div on the category icon

PR #7435, opened 2020-08-11
[K5.1] Custom images for Categories not working properly #5095

xillibit:
Pull Request for Issue #5095  . 
 
#### Testing Instructions

### DIFF
--- a/src/libraries/kunena/icons/icons.php
+++ b/src/libraries/kunena/icons/icons.php
@@ -1356,7 +1356,7 @@ class KunenaIcons
 
 		if ($topicicontype == 'image')
 		{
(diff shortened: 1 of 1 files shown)


rich20:
It works, but only if is chosen with Crypsis the option bootstrap 2 icons and for Crypsisb3 bootstrap 3 icons.
If you use legacy symbols, you need for all categories own images, otherwise the other categories have no symbols.

Bildschirmfoto von 2020-08-11 16-20-19
------------------
With Font awesome (but that's another problem and has nothing to do with this fix) you see additional the flashing question mark.

Bildschirmfoto von 2020-08-11 16-49-32

xillibit:
I will re-do it on 5.2

PR #7455, opened 2020-08-19
[K5.2] Fix issue that js files not copied during install and clean them

xillibit:
Pull Request for Issue # . 
 
#### Summary of Changes 
 
The JS files under media/kunena/core/js wasn't copied during Kunena install and the templates assets wasn't cleaned

#### Testing Instructions


rich20:
The path now exists, but there are only 13 js files in the folder media/kunena/core/js
Should be 51 files and the folder "locales" with the datepicker js

xillibit:
Why 51 files ? These 13 files are the same in crypsis, crypsisb3 and crypsis

PR #7666, opened 2020-11-10
[K5.2] Limit allowed BBCode for the new users (no urls, images) #278

xillibit:
Pull Request for Issue #278.
 
#### Summary of Changes 
 
When setting enabled in Kunena configuration, it removes links on new message, edit and signature in profile only if the user has less posts thant defined in Kunena configuration

#### Testing Instructions

PR #7675, opened 2020-11-14
[K5.2] Limit allowed BBCode for the new users (no urls, images) #278

### LINKED ISSUE #278: Limit allowed BBCode for the new users (no urls, images)
See:
http://www.kunena.org/forum/119-feature-requests/20773-limit-posts-by-new-users-to-prevent-spam?limit=10&start=10#114149

xillibit:
Pull Request for Issue #278. 
 
If needed close #278

### DIFF
--- a/src/administrator/components/com_kunena/language/en-GB/en-GB.com_kunena.views.ini
+++ b/src/administrator/components/com_kunena/language/en-GB/en-GB.com_kunena.views.ini
@@ -612,6 +612,10 @@ COM_KUNENA_DISABLE_RE="Disable Re: on subject"
 COM_KUNENA_DISABLE_RE_DESC="When you enable the option, the Re: will be hidden on the subject."
 COM_KUNENA_CONFIG_DISPLAY_FILENAME_ATTACHMENT="Display attachment filename"
 COM_KUNENA_CONFIG_DISPLAY_FILENAME_ATTACHMENT_DESC="Display attachment filename next to the image or file in messages"
+COM_KUNENA_ADMIN_CONFIG_PREVENT_NEW_USERS_POST_URL_IMAGE = "Set if you want that new users are able to add url and image in his messages or profile"
+COM_KUNENA_ADMIN_CONFIG_PREVENT_NEW_USERS_POST_URL_IMAGE_DESC = "Set if you want that new users are able to add url and image in his messages or profile description. You need to define with the other setting the number minimum of user posts."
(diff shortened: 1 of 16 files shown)


xillibit:
Improvements to be made, see the old PR : https://github.com/Kunena/Kunena-Forum/pull/7666#issuecomment-725477128

xillibit:
@rich20 : can-you try now ?

rich20:
- [ ] If the option enabled, the normal user cannot use BBCode at all such as bold, confidential and so on (also not when the limit is reached).
- [ ] Wrong message: Your message has been successfully posted. (there should be a hint when BBCode has been removed)
- [x] Links added with BBcodes, and also entered links without code will removed (it remains only the brackets)

rich20:
@xillibit If a link has been removed it remain two brackets instead of the link. () Should be there the hint in it?

xillibit:
It does when you just putting a link in message like that  : 

image

?

In this case it just remove the link and i don't see the two brackets ()

P.S.: if i use a link like this one `https://www.kunena.org/forum/kunena-latest-m-archive/156737-php-7-3-and-leatest-post-module-is-show-nothing?start=10#214381` when i submit the message it remains the `?start=10#214381`

xillibit:
With the last change it is able to find too this kind of url : `https://www.kunena.org/forum/kunena-latest-m-archive/156737-php-7-3-and-leatest-post-module-is-show-nothing?start=10#214381`

xillibit:
The last update is able to recognize this kind of url too : `https://www.kunena.org/forum/kunena-latest-m-archive/156737-php-7-3-and-leatest-post-module-is-show-nothing?start=10`

rich20:
>In this case it just remove the link and i don't see the two brackets ()
With the last change it is able to find too this kind of url: `https://www.kunena.org/forum/kunena-latest-m-archive/156737-php-7-3-and-leatest-post-module-is-show-nothing?start=10#214381`

Depending on the length or type of the url you get different results.
Example: I've entered 2 link addresses with and without BBCode. You see on the second image the result.
Bildschirmfoto von 2020-11-15 16-16-39
Bildschirmfoto von 2020-11-15 16-16-55

xillibit:
In this case i can't post the message : 

image

rich20:
>In this case i can't post the message :

I've cleared all caches, now I can reproduce your message. That's not an error, because the message contains not another text and without text can't be saved.
I've tested also the other codes such as confidential, bold and so on. They're working correctly now. Indeed, I found another issue, but I will create an own ticket for it.
So I mark this pull request as done.

- [x] Success

PR #7733, opened 2020-11-29
[K5.2] Update readme : update joomla version and add php 8 on install

xillibit:
Pull Request for Issue # . 
 
If needed close #


src/install.php
```diff
@@ -27,6 +27,7 @@ class Pkg_KunenaInstallerScript
 	 */
 	protected $versions = array(
 		'PHP'     => array(
+			'8.0' => '8.8.0',
```
810, line 30: typo 8.0.0

PR #7771, opened 2020-12-19
Veekee fr topic icon crypsis

VeekeeFr:
Issue
When displaying topic or creating a topic, on a crypsis theme-based forum, the topic icons are shown as large as possible.
image

#### Summary of Changes

Forcing topic icon to match the element style it's included on.

#### Testing Instructions

Check topic and topic creation display.


xillibit:
Can-you check please if the fix should be applied too on crypsisb3 and crypsisb4 ?

PR #7882, opened 2021-01-29
[K5.2] Add check for php extensions during Kunena installation

xillibit:
Pull Request for Issue # . 
 
If needed close #


xillibit:
@rich20 : can-you test it ?

rich20:
I see no error when I install Kunena, but have it not yet tested when some extensions are not installed. At the moment I've trouble with my PC and must solve this first.

rich20:
In principle, it works, but only for extensions, which also are listed in line 54
`	protected $extensions = array('dom', 'gd', 'json', 'pcre', 'SimpleXML');`

**That's an example result:**
 Notice
Required PHP extension 'json' is missing. Please install it into your system.
×
Warning
Extension Update: Custom install routine failure.
Package Update: There was an error installing an extension: com_kunena_v5.2.3-DEV.zip
×
Error
Error installing package

Issue #7945, opened 2021-03-07
The menu item Recent Topics shows either all or nothing


rich20:
**Describe the bug**
Choose all the categories that you want to show or hide in recent topics works not as it should. If you set a single category to show, it shows all categories. Vice versa, if you set a category to hide, then all categories are hide.
It's indifferent whether you set this in the configuration or in the menu item (see the example images).

**Screenshots**
![Bildschirmfoto von 2021-03-07 19-07-53](https://user-images.githubusercontent.com/888847/110250476-af149a00-7f7b-11eb-8e25-e67ed9f2e2cc.png)<br>
![Bildschirmfoto von 2021-03-07 19-08-58](https://user-images.githubusercontent.com/888847/110250485-bb98f280-7f7b-11eb-8a7e-9b38ea4466e7.png)

**System information (please complete the following information)**
Joomla version: 3.9.24 and 3.9.26-dev
Kunena version: 5.2.2 and 5.2.3-dev
Kunena template name and version: all our templates
Php version: 7.4.14
Database version:

PR #7975, opened 2021-03-24
[K5.2] The menu item Recent Topics shows either all or nothing #7945

xillibit:
Pull Request for Issue #7945. 
 
If needed close #
 
#### Testing Instructions

@rich20 : can-you give a try before i merge it ?

After install the build with this fix, in tab frontend of Kunena configuration, if the setting "Category List in Recent Topics" wasn't set, you should have that now : 

image

Try to change the setting "Category List in Recent Topics" and check if it behave like it should on recent discussions

In Kunena menu, it's the same :  
image

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [5.2.4-dev]
 ### Fixed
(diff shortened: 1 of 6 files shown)


xillibit:
My following fix that i had done before has broke this feature : https://github.com/Kunena/Kunena-Forum/pull/7728

Issue #8166, opened 2021-06-09
[CKEditor] CKEditor bug using lists


rich20:
**Describe the bug**
https://www.kunena.org/forum/k-5-2-support/161828-ckeditor-bug-using-lists

**To Reproduce**
Open the editor and choose the Source mode
Enter a code like this:
```
[b]header[/b]
[list]
[*]line 1
[*]line 2
[*]line 3
[/list]
```
Now switch several times between Source and Normal mode and you see the problem

**Actual result**
It adds new code tags every time and messes everything up. The same occur if you want to edit a message with a list.

**Screenshots**
![Bildschirmfoto von 2021-06-09 16-06-02](https://user-images.githubusercontent.com/888847/121371146-9da5aa00-c93d-11eb-9154-e4effab14438.png)

**Additional context**
The same does not happens on the Test-CKEditor, maybe it's a conflict with one of our plugins.
https://ckeditor.com/docs/ckeditor4/latest/examples/bbcode.html

810:
I can't reproduce this issue. (tested on K6.0)

Ruud68:
just ran into this issue with [url]...[/url]
so now it is also impossible to edit a post with an url in it

@810 the issue is on 5, not 6. maybe there is some hint as to why it is working correct on 6 and not on 5?

baticag:
This editor or implemention of it is shit... Errors with every bbcode tag... it's duplicating them, pushing them to the end. Please fix this issues or give us the option back to change to classic editor.

Ruud68:
when reverting this PR  #8115 (so basically reverting ckeditor.js back to version 4.16.0) everything is working again.
I cannot reproduce this issue on the ckeditor website :(
Is kunena using a custom build of ckeditor 4?

Ruud68:
ping @xillibit for relation to #8115

Ruud68:
So for any body for whom this is a blocking issue (me :S).
make a backup of ./media/kunena/core/js/ckeditor.js
then replace the contents of ckeditor.js with this one: https://raw.githubusercontent.com/Kunena/Kunena-Forum/a5fe11dfed4795e70f6b946d125a9090a62d6e6a/src/media/kunena/core/js/ckeditor.js

This will revert the editor version back to 4.16.0 that doesn't have this issue.
Note this is a workaround and may re-introduce other issues solved in (custom) 4.16.1

rich20:
>I can't reproduce this issue. (tested on K6.0)

[...]

PR #8202, opened 2021-06-29
[K5.2] [CKEditor] CKEditor bug using lists #8166

xillibit:
Pull Request for Issue #8166. 
 
If needed close # 
 
#### Testing Instructions

@Ruud68 : you can try this PR


Ruud68:
Okay, so i tried this PR and I can confirm that it is working correct:
I have created a post with ordered / unordered lists, url and code
toggle source code (multiple times) doesn't break the tags
saving the topic and editing also doesn't break the tags.

I have opened an older post (with old list style: [ul][li]...[/li]
this displays not as list but as code in the editor, saving when saving and viewing the post it shows correct as list.

So this is good news :)

PR #8380, opened 2021-09-05
[K6.0] When ban an user, it blocks itself in Joomla!

xillibit:
Pull Request for Issue # . 
 
If needed close #

### DIFF
--- a/src/libraries/kunena/src/User/KunenaBan.php
+++ b/src/libraries/kunena/src/User/KunenaBan.php
@@ -811,7 +811,8 @@ class KunenaBan extends parentAlias
 
 		if ($this->userid)
 		{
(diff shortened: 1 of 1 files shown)


src/libraries/kunena/src/User/KunenaBan.php
```diff
@@ -811,7 +811,8 @@ public function save($updateOnly = false): bool
 
 		if ($this->userid)
 		{
-			$user = Factory::getApplication()->getIdentity($this->userid);
```
mahagr, line 814: Yeah, the above is supposed to get the current user. The parameter is for initialization only.

xillibit: The issue is that the method getIdentity() doesn't accept any parameters

PR #8701, opened 2022-01-01
[K6.0] Try to fix undefined constant KPATH_ADMIN in plugin kunena_joomla

xillibit:
Pull Request for Issue # . 
 
If needed close #

### DIFF
--- a/src/plugins/plg_kunena_joomla/joomla.php
+++ b/src/plugins/plg_kunena_joomla/joomla.php
@@ -35,7 +35,7 @@ class plgKunenaJoomla extends CMSPlugin
 	{
 		parent::__construct($subject, $config);
 
-		$this->loadLanguage('plg_kunena_joomla.sys', JPATH_ADMINISTRATOR) || $this->loadLanguage('plg_kunena_joomla.sys', KPATH_ADMIN);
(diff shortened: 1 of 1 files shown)


woluweb:
Hi,
I had a similar issue with Kunena when using Community Builder.
See details below

-----------

Blank page in the front-end because of plugin Kunena - Community Builder

When enabling PHP8.0 or PHP8.1 would cause a blank page in the front-end, without any indication even with Error Reporting set on Maximum and Debuggin enabled.

When switching to Protostar we could see an error message: undefined constant KPATH_ADMIN

After some search, we discovered that renaming the plugin/kunena/comprofiler folder would solve the issue.

Then we saw that this comprofiler plugin was actually Kunena - Community Builder Integration.

Given the fix mentioned here https://github.com/Kunena/Kunena-Forum/pull/8701/commits/71fdc565cf9aab5f21df0094d2bbc7bc577ade94

we edited /plugins/kunena/comprofiler/comprofiler.php and applied the same fix on line 72, namely changing

$this->loadLanguage('plg_kunena_joomla.sys', JPATH_ADMINISTRATOR) || $this->loadLanguage('plg_kunena_joomla.sys', KPATH_ADMIN);

by

$this->loadLanguage('plg_kunena_joomla.sys');

PR #8724, opened 2022-01-07
fix ckeditor not respecting configured buttons

Ruud68:
Pull Request for Issue:
Template ckeditor button settings not handled correct.
If you e.g. remove bold (set to no), then it still displays
 
#### Testing Instructions
Install PR, change buttons in template, see if button toggles in editor


xillibit:
Tested with CKeditor and bbcode, it's ok

PR #8877, opened 2022-03-29
[Bugfix] Fix stripped/broken [code] bbcode when "type" is not specified

fevangelou:
When [code] blocks have no specified "type" (e.g. for php, js etc.), the resulting HTML to be rendered is `<pre xml:>...</pre>` which in turn breaks the final HTML that is output by Kunena. This simple check ensures the validity of the HTML output to be rendered.

Resolves the issue reported originally here: https://www.kunena.org/forum/k-5-2-support/161931-bug-in-code-code-parser


fevangelou:
Thank you!

PR #8995, opened 2022-05-13
Refactor closable alerts to BS5 syntax

Ruud68:
Pull Request for Issue # . 
 
#### Summary of Changes 
refactored closable alerts to BS5

#### Testing Instructions
category description, topic headerdesc and announcements should display okay and conform BS5 alert. They should be closable via X button

Note: haven't tested announcement change myself

### DIFF
--- a/src/site/template/aurelia/layouts/category/item/default.php
+++ b/src/site/template/aurelia/layouts/category/item/default.php
@@ -28,9 +28,9 @@ $this->addStyleSheet('rating.css');
 <?php if ($this->category->headerdesc) : ?>
     <div class="clearfix"></div>
     <br>
-    <h1 class="alert alert-info shadow-lg rounded">
-        <a class="close" data-bs-dismiss="alert" href="#"></a>
+    <h1 class="alert alert-info shadow-lg rounded alert-dismissible fade show">
 		<?php echo $this->category->displayField('headerdesc'); ?>
+		<button type="button" class="btn-close" data-bs-dismiss="alert" aria-label="Close"></button>
     </h1>
 <?php endif; ?>
 
--- a/src/site/template/aurelia/layouts/topic/item/default.php
+++ b/src/site/template/aurelia/layouts/topic/item/default.php
@@ -56,9 +56,9 @@ if ($this->topic->locked)
 ?>
 <div class="kunena-topic-item <?php echo $txt; ?>">
 	<?php if ($this->category->headerdesc) : ?>
-        <div class="alert alert-info shadow-lg rounded">
-            <a class="close" data-bs-dismiss="alert" href="#">&times;</a>
+        <div class="alert alert-info shadow-lg rounded alert-dismissible fade show">
 			<?php echo $this->category->displayField('headerdesc'); ?>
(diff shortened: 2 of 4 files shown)


xillibit:
To close with X button it has no effect on category herder description and announcements, not tried topic headerdesc

Ruud68:
With which Joomla template are you testing?

xillibit:
The default Joomla! template Cassoepia

Ruud68:
Okay, had a look at this and would like to welcome you to Cassiopea.
The cassiopea template by default doesn't load bootstrap as a whole, but loads bootstrap in parts. So instead of loading bootstrap.bundle.js it will load the individual js files it needs for specific views. The loading needs to be done per view.
From a performance perspective I understand why it is implemented this way, but for an extension developer this adds to the complexity.
This means that you need to use default bootstrap alerts in Kunena you need to add it in the views that display these alerts by adding the code HTMLHelper::_('bootstrap.alert');

But what if the user is not using casseopia but another bootstrap 5 template? Then most likely alert.js functionality is already loaded via bootstrap.bundle.js leading to unpredictable / broken behavior.

That is exactly the reason why I readded the load bootstrap css parameter in aurelia: Kunena aurelia by default loads bootstrap from Joomla, but when you are using a bootstrap 5 template then it will load the bootstrap functionality twice leading to a broken site (the JS functions do not work anymore)

Because cassiopea doesn't fully load bootstrap, Kunena needs to be aware of that and cannot use the full potential of bootstrap. or it needs to load these script manually but giving the user an opt-out via a config setting because when the user has full-blown bootstrap template this will likely break.

I have added it to the aurelia template, so now the alerts should work.

xillibit:
I'am testing only with Joomla! default template, now with your changes it working

Maybe in this case we should detect in Kunena which Joomla! template is used and if it's Cassoepia loads bootstrap differently

Ruud68:
This needs some thought. I am developing a full blown bootstrap5 as it was intended template :) so not stripped / biased.
That is also the reason why I am currently developing a Kunena template that uses 100% bootstrap without the user needing to worry / think why things are not working etc.
The difficulty here is that cassiopea is not implementing full bootstrap but does it in parts where you need to think what parts you need. For extension devs it would be easier to if that wasn't needed as that would then automatically make their extension work on all BS5 templates.

So you either add a config parameter where you toggle between everything bootstrap related: load (because kunena is responsible for loading additional BS scripts / css) or not load (because the site template is loading everything needed)
This way the user can (in case off issues use this toggle.
This is what we have now (with the (wrongly named) load bootstrap css.

Or you implement an automated toggle where the toggle logic is cassiopea (load) or not (not load)

I think the manual toggle gives more control: e.g. in case of a non-cassiopea template that implements also in parts (via joomla build in bootstrap instead of template delivered bootstrap)

Ruud68:
and... a toggle would make more sense because you need to load when using Kunena on a non BS5 template.
So have Kunena have  a parameter to load bootstrap (what you have now) or not load bootstrap (because the template is loading everything needed).
You then only need to rename the 'load Bootstrap CSS' in the template to 'load Bootstrap'

Also everything bootstrap loading related should be done in the (aurelia) template and not in Kunena itself (not sure if this is the case)

PR #9014, opened 2022-05-17
Working draft

Ruud68:
Pull Request for Discussion https://github.com/Kunena/Kunena-Forum/discussions/9012 
 This is a proposal, let me know what you think :)

#### Summary of Changes 
the 'Load Tooltips' toggle in aurelia will add (via JS) bootstrap tooltips (if yes) or remove all tooltip (if no) including default browser tooltips.

It will handle all elements that are in #kunena 'container' and have a title attribute (The title attribute is used to display the tooltip in the browser).

The class .hasTooltip is not used anymore other then that it can be used for styling

#### Testing Instructions
toggle 'Load Tooltip' to no and check that there are not tooltips displayed
toggle 'Load Tooltip' to yes and verify that elements have a bootstrap tooltip (so not a browser default tooltip)


xillibit:
I have this error in javascript console of Firefox, when going to a topic : 

image

Ruud68:
@xillibit what loads dom.js? I am loading the toolltips conform bootstrap documentation, It looks like dom.js loads after kunena loading the tooltips and tries to load them 'again'?

xillibit:
It loaded from http://joomla-41-dev-kunena-60/media/vendor/bootstrap/js/dom.js?5.1.3

Ruud68:
Okay so I found the issue. it is related to this: https://github.com/twbs/bootstrap/issues/33940
in tooltips.js I am selecting all elements that have a title attribute, but a dropdown or popover also has a title attribute. So when that happens, the tooltip is assigned to the element and then the popover or dropdown tries to instantiate but can't because there can only be one.

Now I can exclude popovers, etc ehn assigning the tooltip, but that will always be error prone especialy because we do not have control over what else is on the page (maybe a carousel, or other element with title).

So this approach is not maintainable, we need to specifically add the attribute data-bs-toggle="tooltip" on all elements and then in the tooltips.js (or main.js) trigger on that as selector.

I will change this PR to do this.

Ruud68:
@xillibit could you retry?

xillibit:
thanks now the error has disappear

PR #9080, opened 2022-06-03
add markup to mail templates

Ruud68:
Pull Request for Issue

mail templates have body consisting of one single line of text (no 'paragraphs / line breaks) 
 
#### Summary of Changes 

did some markup to the body of the mail send via the MailTemplate 

#### Testing Instructions
create topic / reply to topic / report message

mail received should be multi line instead of one line containing all content.


Leen29:
then it looks like this is that ok
isthissogood
?

Ruud68:
That is correct @Leen29  the body now has multiple lines instead of all information on one long line

Issue #9113, opened 2022-06-15
Guests allowed to view images / attachments even if set to NO


pippo8765:
Parameter Allow images / attachments for guests set to No, but they ARE allowed to view images / attachments in post as well as clickable thumbs in post bottom. Template Aurelia.

To reproduce just set this to No and attach image in topic. Do not login and visit topic.

**Expected behavior**
Dissallowed view.

Joomla version: 4.1.4
Kunena version: 6.0
Php version: 8.x

Ruud68:
@pippo8765 can you test https://github.com/Kunena/Kunena-Forum/pull/9117 ?

rich20:
- [x] Success

PR #9117, opened 2022-06-16
fix showing images / attachments to guest when configured to not display

Ruud68:
Pull Request for Issue #9113  . 
 
#### Summary of Changes
reverted return on functions authoriseRead/Own/Exists()
not for this specific issue (images and attachments showing for guests when not allowed to show) only the authoriseRead return should have changed, but I figured that this issue is also valid for authoriseOwn and authoriseExists
 
#### Testing Instructions
create message with image and or attachment
toggle 'Show Images for Guests' and 'Show Attachments for Guests' and see if the image / attachment display / do not display correct

### DIFF
--- a/src/libraries/kunena/src/Attachment/KunenaAttachment.php
+++ b/src/libraries/kunena/src/Attachment/KunenaAttachment.php
@@ -767,7 +767,12 @@ class KunenaAttachment extends KunenaDatabaseObject
 
 				try
 				{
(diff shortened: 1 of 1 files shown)


pippo8765:
Seems ok now! I'll go with this solution for now, works for my case.

Leen29:
This seems to work well on my test sites too

Issue #9123, opened 2022-06-19
The Kunena menu type "Create New Topic" has no selection options


rich20:
**Describe the bug**
https://www.kunena.org/forum/k-6-0-support/165212-k6-0-0-create-a-subject,-no-more-category-option
Confirmed.
The Kunena menu type "Create New Topic" has no selection options. It is not possible to select a category in which topics should be created by default.

**System information**
- Joomla version: 4.1.5-rc3-dev
- Joomla template: CASSIOPEIA
- Kunena version: 6.0.1-DEV
- Kunena template: Aurelia
- Php version: 8.0.16
- Database version: 10.5.15-MariaDB

**Desktop**
- OS: Ubuntu 22.04 LTS
- Browser: Chromium, Firefox and Opera

Leen29:
Hi Rich,
I think it's about the adminstrator section when creating a new kunena menu item category-list, there you could previously choose which category should be displayed.
and also create a new topic
![k5categorielist](https://user-images.githubusercontent.com/17086048/174483771-23d7239d-aa4e-40ee-a1a4-dc208c808219.jpg)
![k5newtopic](https://user-images.githubusercontent.com/17086048/174483777-3168645c-d62d-4e9c-b000-2857cb9db5bb.jpg)
![k6newtopic](https://user-images.githubusercontent.com/17086048/174483782-7b2a7671-d6b4-4ddf-acce-3194b94767c9.jpg)
![k6categorielist](https://user-images.githubusercontent.com/17086048/174483783-fe1bedb5-8d2a-4135-b758-b34c1959cdbe.jpg)
c

Leen29:
I forgot to mention that if you create a menu item topic, this one at the front then suddenly gives the option to choose category
![K6frontmenuitem](https://user-images.githubusercontent.com/17086048/174484152-7179f803-1a96-47fe-b75d-0a8e2ef81496.jpg)
.

rich20:
@Leen29 With this option it will create the topics in the selected category, regardless of how the categories are sorted.

------------------------------------------------------

Works now

PR #9124, opened 2022-06-19
[K6.0] Kunena menu type "Create New Topic" no selection options #9123

xillibit:
Pull Request for Issue #9123. 
 
#### Testing Instructions

### DIFF
--- a/src/site/tmpl/topic/create.xml
+++ b/src/site/tmpl/topic/create.xml
@@ -7,7 +7,7 @@
 
     <!-- Add fields to the request variables for the layout. -->
     <fields name="request">
-        <fieldset name="request">
+        <fieldset name="request" addfieldprefix="Kunena\Forum\Administrator\Field">
             <field name="catid"
                    type="kunenacategorylist"
                    default=""


Leen29:
I had tested the last change but it didn't work for me yet on test site.
after some searching it turns out that in my version the list.xml was missing in
\components\com_kunena\tmpl\category\
if I add it from the github version it works
so think this one is still missing in pkg_kunena_v6.0.0_2022-06-1
voorbeeld
0

Ruud68:
Hi @Leen29 
the fix is not in the downloaded file you are using as the fix was after that file was created later. You need to wait for 6.0.1 or build a file yourself.

Leen29:
Yes Ruud thank you   I already understood that but then I noticed that
so all that change was not enough because the
"list.xml was missing" in testwebsite\components\com_kunena\tmpl\category"
and when I put it in another local test site it worked well.
so I was a bit surprised since that was not mentioned in the adaptation
but i am only a kunena hobbyist so maybe I don't understand everything very well sometimes

PR #9341, opened 2022-10-31
Fix: missing namespacing in newly added (email) files

Ruud68:
Pull Request for Issue # . 
 
#### Summary of Changes
no namespacing set in newly (re) added email files
 
#### Testing Instructions


Leen29:
I made that namespacing set in newly (re) added email files on a test site and everything seems to be working fine.
But may I ask why that had to what does it do.?

Ruud68:
@Leen29 Namespacing is an advanced PHP feature that allows for e.g. autoloading of files etc. It works without the namespacing being set because the namespacing is also set on a higher level, but it can lead to issues when used differently. That is why this PR to make it conform standard.

Leen29:
@Ruud68
Thanks for the explanation, I learned something new

Issue #9329, opened 2022-10-28
custom topic items not working


Ruud68:
**Describe the bug**
when creating custom topic items these are not working correct (images not found)
**To Reproduce**

1. copy directory + contents ./media/kunena/topic_icons/default to ./media/kunena/topic_icons/test
2. select a Kunena category in the back-end and set 'test' as topic icons (instead of default)

**Expected behavior**
When browsing the category, the icons should be okay

**Actual result**
icons (svg) are taken from ./media/kunena/core/svg/test/[filename]

which is not existing as the icons for this set (and for the default set as well!!) are in ./media/kunena/topic_icons/test

Joomla version: 4.2.4
Kunena version: 6.0.4-dev
Php version: 8.0

rich20:
Confirmed
It shows this error:
Warning: DOMDocument::load(): I/O warning : failed to load external entity "/.../media/kunena/core/svg/own_symbols/user/default.png/user/file.svg" in /.../libraries/kunena/src/Icons/KunenaSvgIcons.php on line 68

Ruud68:
This is something I never got my head around in Kunena.
The topic icons are dependent on the labels.xml. That can be changed via adding a labels.xml to the template config directory. Never understood how these interacted, now I am migrating a site that is actually using topic icons and custom labels and can't seem to get it working.
Is there some sort of documentation on how to set that up? If I understand how it works (without wrestling though all the code to reverse engineer it, then maybe we can discuss how to change / fix as we now have an opportunity as what was working in K5 is not migrated over to K6

rich20:
Topics icons and labels are two different things. 
The labels.xml can be copied to /components/com_kunena/template/aurelia/config/ and edited. The symbols you can change in it only affect the labels themselves, if you use Fontawesome (see image).
![image10](https://user-images.githubusercontent.com/888847/198710984-b204416c-0aae-4a36-9a65-57e7252fc151.png)

In principle you need only two changes.
Example:
id="0" name="COM_KUNENA_LABELS_QUESTION" b2="question-sign" b3="question-sign" fa="question-circle" labeltype="primary"

id="0" name="**your definition**"  b2="question-sign" b3="question-sign" fa="car" labeltype="**secondary**"
If you change the label type, you get a separate class

```
.badge.me-1.bg-secondary {
  background: green !important;
}
```

The topic symbols can be changed in the topicicons.xml. To do, you just had to copy and rename the folder "default" and edit the contained topicicons.xml.
[...]

PR #9342, opened 2022-10-31
fixed topicicons + svg icons on labels

Ruud68:
Pull Request for Issue #9329 . 
 
#### Summary of Changes

- topic_icons default where using svg files from media/kunena/core/svg and not the ones in media/kunena/topic_icons/default/[user|system]/[svgiconname]
- custom topic_icons where not loading svg icons
- media/kunena/core/svg is now set as fallback when svg icon configured is not found (this because all used svg icons in default topic_icons are not in hte user / system svg directory
- labels.xml now has possibility to also use svg icons (next to bs3 and fa)

 
#### Testing Instructions

- install PR and check if all icons (svg) are displaying as before
- add custom topic_icons set and check if icons are correct as configured
- load default labels.xml and check if when svg is configured, svg icons are loaded and displayed
-  create labels.xml in template config directory and configure svg icons

### DIFF
--- a/src/libraries/kunena/src/Icons/KunenaSvgIcons.php
+++ b/src/libraries/kunena/src/Icons/KunenaSvgIcons.php
@@ -52,11 +52,19 @@ class KunenaSvgIcons
 		}
 		elseif ($group == 'systemtopicIcons')
 		{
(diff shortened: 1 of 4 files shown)


Leen29:
Files on testsite modified and tested and works.

Issue #9344, opened 2022-10-31
Error when Icon Type = SVG, Selectable Topic Icons = No and labels are active


rich20:
**To Reproduce**
Aurelia setting - Icon Type = SVG, Show Labels = Topic id or Custom id
Configuration (tab) Frontend - Selectable Topic Icons = No
Go in frontend to Recent Topics

**Actual result**
Warning: DOMDocument::load(): I/O warning : failed to load external entity "/.../media/kunena/core/svg/comments.svg" in /.../libraries/kunena/src/Icons/KunenaSvgIcons.php on line 76

**System information**
- Joomla version: 4.2.5-rc2-dev
- Joomla template: CASSIOPEIA
- Kunena version: 6.0.5-DEV
- Kunena template: Aurelia
- Php version: 8.0.20
- Database version: 10.5.16-MariaDB

**Desktop**
- OS: Ubuntu 22.04.1 LTS
- Browser: Chromium, Firefox and Opera

rich20:
I see now, it has nothing to do with the labels. The comment.svg do not exist in this path /media/kunena/core/svg/**comments.svg**

rich20:
Solved

PR #9345, opened 2022-11-01
add missing svg icons

Ruud68:
Pull Request for Issue #9344 . 
 
#### Summary of Changes 
copy over missing svg icons from ./media/kunena/core/svg/default/[user|system]/svg to ./media/kunena/core/svg as that is used as fallback location
 
#### Testing Instructions
@rich20


Ruud68:
@xillibit when looking at the comments.svg, they look good on the recent topics overview (smal in first column), but when viewing on the topic page itself, the icon is huge.
I noticed that the size of the icons I copied over in this PR are different in size, so when no size is provided they are displayed big.
maybe @rich20 can confirm?

This I do not get because on the whole kunena installation there is only one comments.svg and that was used previously on these pages and that is the one I copied over.

Do you know how we can get these files in the same size as the ones already in the svg directory (so the comments.svg)?

rich20:
For me, the symbols are displayed correctly in all places. I have also tested it with a self-created svg image (size 450 x 530), which is also displayed correctly everywhere.

Ruud68:
did you test with our without #9348
That has been merged now and that fixes it for for me

rich20:
I had tested it before this fix, but had already uploaded the icons manually when I saw this error message.

> Warning: DOMDocument::load(): I/O warning : failed to load external entity "/.../media/kunena/core/svg/comments.svg" in /.../libraries/kunena/src/Icons/KunenaSvgIcons.php on line 76

Now is all ok!

PR #9349, opened 2022-11-02
attachments cart is not card

Ruud68:
Pull Request for Issue #- . 
 
#### Summary of Changes 
attachments in posts are displayed in a bootstrap card, but because it has wrong class name (cart !== card) it doesn't display as BS card.
This PR changes that.
Note that this is a visual change!
 
#### Testing Instructions


Ruud68:
I noticed on my own template that when all attachments are displayed inline (in the message), the the attachment card is shown (empty).
I have changed that to only show the attachment card when there is actually something to show: so when there are attachment, they are readable and they are not already displayed in the message it self.

Issue #9406, opened 2022-11-30
No mail sent to moderator when moderator email == forum email (part 2)


rich20:
**Describe the bug**
https://github.com/Kunena/Kunena-Forum/issues/9401

This fix is only a half-solution. It only works with the additional security setting "E-mail Moderators = All New Posts". 
Example: 
1. leave the default setting "E-mail Moderators = Unapproved Posts".
2. subscribe to a category with the global moderator
3. open a new topic or post a reply with a test user

**Result**
The moderator does not receive a subscription mail

**System information**
- Joomla version: 4.2.6-dev
- Joomla template: CASSIOPEIA
- Kunena version: 6.0.7-DEV
- Kunena template: Aurelia
- Php version: 8.0.20
- Database version: 10.5.16-MariaDB

**Desktop**
- OS: Ubuntu 22.04.1 LTS
- Browser: Chromium, Firefox and Opera

PR #9407, opened 2022-11-30
moderator unapproved posts new topic

Ruud68:
Pull Request for Issue #9406  . 
 
#### Summary of Changes 
it will now also handle global moderators that are subscribed to a topic and that have the same email as the site
 
#### Testing Instructions
@rich20


rich20:
It works as it should, if you enter a value higher than 0 for email to multiple recipients. I think that's ok, because there's no point in entering a valid email address in this field if this option is not used.

Issue #9404, opened 2022-11-29
Tooltip CSS


LinelabCZ:
When I disable toolitps in the Aurelia template settings the user type CSS class disappears.

enable
`<span class="kwho-admin hasTooltip">`

disable
`<span class="">`

The tooltip now appears above the image directly in the forum messages.
```
<a href="[/forum/attachment/3874](view-source:/forum/attachment/3874)"
       data-bs-toggle="tooltip" title="...a4f66e1a03e3ba" data-fancybox="gallery">
```
System information

Joomla version: 4.2.5
Kunena version: k6.0.6

Ruud68:
I don't know how to reproduce, not seeing this issue (but then again, not sure where to look).
Can you give a view where you see this, can you tell if image is attachement or inserted in message, can you make a screenshot of where you see the title when tooltips are turned off?

rich20:
@Ruud68 Go to the index page
User colour with activated tooltips in the Aurelia setting
.
![image1230](https://user-images.githubusercontent.com/888847/205443433-5d103240-8622-47c4-b374-d015a2deffe5.png)
.
User colour with disabled tooltips in the Aurelia setting (registered user sees link color)
.
![image12220](https://user-images.githubusercontent.com/888847/205443641-bb1401d7-5067-4ca1-a7d4-998cfae76ddd.png)
.

User colour with disabled tooltips in the Aurelia setting (guests sees link hover color)
.
![image1253](https://user-images.githubusercontent.com/888847/205443700-515e216d-383e-4151-91fa-1b96c16c6ae4.png)

.

PR #9414, opened 2022-12-04
fix tooltip on user name whe tooltips = disabled

Ruud68:
Pull Request for Issue #9404 . 
 
#### Summary of Changes 
class is automatically added to user when class = null, but when tooltips = disabled, then class <> null but is empty > therefore the class is not automatically added

#### Testing Instructions
install PR and check if issue is fixed: also check all other views where user name is displayed


Ruud68:
pinging [@linelabcz](https://github.com/LinelabCZ) and @rich20

rich20:
- [x]  Success

Issue #9419, opened 2022-12-11
On index page when there are sub-categories there are displayed vertically


xillibit:
The sub-categories there are displayed vertically :

![image](https://user-images.githubusercontent.com/876395/206932062-bb900158-5b93-4007-b006-eebbd7af6a90.png)

Expected result the same as Kunena 5.2 : 

![image](https://user-images.githubusercontent.com/876395/206932105-da1fd94f-d15e-4b9b-805d-d326a94bb1f1.png)

xillibit:
@Ruud68 : do-you have an idea how to fix this one ?

Ruud68:
I'll check this tomorrow, wanted to check earlier but dropped of my screen due to long todo list :)

PR #9421, opened 2022-12-14
Make list of subcategories inline

Ruud68:
Pull Request for Issue #9419 . 
 
#### Summary of Changes 
class changes to make sub categories float / inline like they were in K5.2
 
#### Testing Instructions
@xillibit


xillibit:
Just tested seems ok

Issue #9422, opened 2022-12-14
Announcement displayed although unpublished


Ruud68:
**Describe the bug**
after upgrade 6.0.6 to 6.0.7 all my 'historic' announcements started to show.
I needed to set the unpubled date to a real date (somehow it was set to 1000-00-00 on upgrade.

Unpublishing the announdement is ignored, see screen captures for where an announcement that is unpublished is still being displayed

**To Reproduce**

**Expected behavior**
announcement publishing should follow publish status set and start / stop publish dates.

**Actual result**
After the upgrade to 6.0.7 announcement start showing

**Screenshots**
![image](https://user-images.githubusercontent.com/2733197/207691179-54fcef7d-3dbf-4408-abbf-1f9b9b651215.png)
![image](https://user-images.githubusercontent.com/2733197/207691275-d169a901-2dc7-4033-a62c-7014657259a8.png)
![image](https://user-images.githubusercontent.com/2733197/207691388-8cf8cdf9-73ea-4104-981c-f5f1513432b3.png)

**System information (please complete the following information)**

Joomla version: 4.2.6
Kunena version: 6.0.7 (started direct after upgrade to this version)
Php version: 8.0.26
Database version: 10.3.37-MariaDB-0ubuntu0.20.04.1

PR #9423, opened 2022-12-15
Fix for announcement query bug

Ruud68:
Pull Request for Issue #9422 . 
 
#### Summary of Changes
Kunena 6.0.7 exposed a bug in the announcement selection logic.

due to a bug in the query, unpublished announcement where displayed when publish up / down conditions where met.
 
#### Testing Instructions
install PR, unpublished announcement should not display, publish up and publish down should be respected.

Important note:
There was a change in the default date/time for publish_up / publish_down. this changed to a default value of '1000-01-01 00:00:00'
an update script changed these values where the date/time was 'null'

But that was not only the use case: date/time can also be the default nullDate being '0000-00-00 00:00:00'
these nulldate values where not replaced resulting in wrong publish_down action.

### DIFF
--- a/src/libraries/kunena/src/Forum/Announcement/KunenaAnnouncementHelper.php
+++ b/src/libraries/kunena/src/Forum/Announcement/KunenaAnnouncementHelper.php
@@ -58,14 +58,14 @@ abstract class KunenaAnnouncementHelper
 
         $id = \intval($identifier);
 
-        if (empty(self::$_instances [$id])) {
-            self::$_instances [$id] = new KunenaAnnouncement(['id' => $id]);
-            self::$_instances [$id]->load();
+        if (empty(self::$_instances[$id])) {
+            self::$_instances[$id] = new KunenaAnnouncement(['id' => $id]);
+            self::$_instances[$id]->load();
         } elseif ($reload) {
-            self::$_instances [$id]->load();
+            self::$_instances[$id]->load();
         }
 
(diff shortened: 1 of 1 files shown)


xillibit:
> But that was not only the use case: date/time can also be the default nullDate being '0000-00-00 00:00:00'
> these nulldate values where not replaced resulting in wrong publish_down action.

In which case it set the date to '0000-00-00 00:00:00' ?

Ruud68:
no idea in which case, can be prior to version 6. Note that 0000-00-00 00:00:00 is the 'official' sql nulldate used by joomla ($db->getNullDate())

main issue for me was that because of the missing parentheses around the publish_up and publish_down 'where' parts, they where treated as OR instead of AND resulting also in published = 0 to be selected (when published_up / down matched)

Ruud68:
from K5.2 installation (prio to updating to K6)
image

xillibit:
> no idea in which case, can be prior to version 6. Note that 0000-00-00 00:00:00 is the 'official' sql nulldate used by joomla ($db->getNullDate())

Since MySQL 5.6 or 5.7 default date has been defined as '1000-01-01 00:00:00.000000' to '9999-12-31 23:59:59.999999', there was a change in Joomla! too at some point

If you keep dates as '0000-00-00 00:00:00'  Mysql 5.6 or 5.7 will give you the following error :  `Incorrect datetime value: '0000-00-00 00:00:00' `

Issue #9424, opened 2022-12-15
Large icons are displayed in sub categories when a category CSS class suffix has been added


rich20:
**To Reproduce**
go in the backend to categories
add to a category which contain sub categories a CSS class suffix like --green
.
![Bildschirmfoto vom 2022-12-15 21-00-09](https://user-images.githubusercontent.com/888847/207956107-498835a5-cf9a-4849-9c6b-66128cf68f0b.png)

**Actual result**
It replaces the small symbols with large ones
.
without CSS class
.
![Bildschirmfoto vom 2022-12-15 21-32-18](https://user-images.githubusercontent.com/888847/207961446-743ed234-2267-4335-8ca3-d33b3782d1ea.png)
.
with CSS class suffix
.
![Bildschirmfoto vom 2022-12-15 21-31-52](https://user-images.githubusercontent.com/888847/207961641-ac84abac-bcc8-489d-9e15-412de3abc3c7.png)

**System information**
- Joomla version: 4.2.6-dev
- Joomla template: CASSIOPEIA
- Kunena version: 6.0.8-DEV
- Kunena template: Aurelia
- Php version: 8.0.20
- Database version: 10.5.16-MariaDB

**Desktop**
- OS: Ubuntu 22.04.1 LTS
- Browser: Chromium, Firefox and Opera

Ruud68:
Hi @rich20 , I did a PR. If you can test that would be great!

PR #9425, opened 2022-12-17
fix svg size of sub category icons when custom class is used on paren…

Ruud68:
…t directory

Pull Request for Issue #9424 . 
 
#### Summary of Changes 
removed 'conflicting' scss

#### Testing Instructions
install PR
visit form (make sure css is regenerated).
test all views and see if icons are still correct (when used in combination with 'category-' class (e.g. category-green)

### DIFF
--- a/src/site/template/aurelia/assets/scss/icons.scss
+++ b/src/site/template/aurelia/assets/scss/icons.scss
@@ -195,7 +195,8 @@ svg {
 .category th svg,
 footer svg,
 .col-md-1 svg,
-.category-stickymsg a svg {
+.category-stickymsg a svg,
+[class*="category-"] th svg {
   width: 34px;
   height: 34px;
 }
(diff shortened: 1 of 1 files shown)


rich20:
- [ ] <strike>Success</strike>

rich20:
Sorry, it was a mistake by me.
It shows on Recent Topics and Recent Posts page now also the mini symbol
.
Bildschirmfoto vom 2022-12-17 12-49-29

Ruud68:
Hi,
~~so what is happening here is that by default the class for this row is "category"~~

~~class category is used to style the row.~~

~~Now by setting "-green" the class is changed to "category-green". This is expected but as category-green is not styled the same as category things are going wrong (like the styling of the svg icon).~~

~~My proposal would be to always set category as class and add category-green additionally when -green is configured.~~

~~but before making this change we can test this by configuring the category class not to "-green" but to "[space]category-green"
where [space] = a space :)~~
~~so now the class will become "category category-green"~~

~~@rich20 can you test if this works? It is used in views:~~

Ruud68:
sorry, never mind the test. did a change to the PR.
Can you test that?

rich20:
- [x] Now everything is correct

PR #9426, opened 2022-12-21
Change: rename user in messages should also rename users in topics

Ruud68:
Pull Request for Issue # . 
 
#### Summary of Changes
currently when syncing usernames in messages, it does just that: sync the username in messages. but the username is also set in topics and there it is not renamed.

This could (and if it can, it will) be a gdpr issue. When a user requests his account to be deleted. The joomla user is renamed.
after approving this, the administrator should sync the usernames in messages to set the new username, but the old username still shows on topics (first post / last post).

This PR changes that: after this PR also the usernames in topics are changed
 
#### Testing Instructions

update usernames and check if they are all replaced in messages, but also in topics.


xillibit:
By what the username will be replaced in this case ?

Ruud68:
with the same name that the username is replaced by in the messages.

the username for someone posting is not only stored in the message, but also in the topic 

image
1 = first_post_guest_name
2 = last_post_guest_name

so when in this caase I change  my name from Ruud van Lent to ABC, then for all messages that are assigned to my user id Ruud van lent will be replaced by ABC, but in the topic overview / recent topics it will still say Ruud van Lent (last_post_guest_name). This PR also changes these two values when applicable to ABC

PR #9444, opened 2023-01-04
Privacy plugin

Ruud68:
Pull Request for Issue # . 
 
#### Summary of Changes 
added new plg_privacy_kunena
changed build script
removed privacy capabilities from system plugin
 
#### Testing Instructions
note: untested as I do not have 6.1 test environment (and installing it would break my current test environment)


src/plugins/plg_privacy_kunena/kunena.script.php
```diff
@@ -0,0 +1,116 @@
+<?php
+
+/**
+ * [___HEADER_PHP___]
+ */
+
+// No direct access
+defined('_JEXEC') or die;
+
+use Kunena\Forum\Plugin\Privacy\Kunena\Helper\OchInstallerScriptHelper;
```
xillibit: Why this class nammed OchInstallerScriptHelper ?

Ruud68: that whole file should not be there (replaced with script.php). It is a 'left over' of my own version of the plugin I developed for testing this. I will remove it :)

src/plugins/plg_privacy_kunena/kunena.xml
```diff
@@ -0,0 +1,30 @@
...
+		https://www.kunena.org</authorUrl>
+	<copyright>www.kunena.org</copyright>
+	<license>GNU/GPLv3 or later</license>
+	<namespace path="src">Kunena\Forum\Plugin\Privacy\Kunena</namespace>
+	<changelogurl>
+		@kunenachangelog@</changelogurl>
+	<scriptfile>script.php</scriptfile>
+	<files>
+		<folder plugin="kunena">
+			src</folder>
```
xillibit: There is a space here before `src` which prevent the plugin to be installed

Ruud68:
Okay, just removed the xml formatting, can you check if it now works correct?

xillibit:
yes install works now

Issue #8945, opened 2022-04-21
Dashboard - the symbols on the labels are not clickable


rich20:
**Describe the bug**
https://www.kunena.org/forum/36-German/164878-dasboard-1-seite-nur-textsensitiv
<br />
![Bildschirmfoto von 2022-04-21 04-32-26](https://user-images.githubusercontent.com/888847/164360626-61d90d29-e654-4b98-a004-487bcc0116b6.png)

**System information**
- Joomla version: 4.1.3-dev
- Joomla template: CASSIOPEIA
- Kunena version: 6.0.0-RC3-DEV-GIT
- Kunena template: Aurelia
- Php version: 7.4.28 and 8.0.13 
- Database version: 10.5.11-MariaDB and 10.4.22-MariaDB

**Desktop**
- OS: Ubuntu 20.04.2 LTS
- Browser: Chromium, Firefox and Opera

rich20:
Solved by this fix: https://github.com/Kunena/Kunena-Forum/pull/9467

PR #9467, opened 2023-02-10
cpanel UX improvement

Ruud68:
Pull Request for Issue # - . 
 
#### Summary of Changes
So last week I was in a zoom session with a user who wanted to show me something in Kunena (6). He was so confused about the cpanel and that the 'title' where clickable, as he saw things moving when hovering the cpanel for e.g. users, but never found where to click.
So I made the whole card clickable as that is what the user expects. it is a BS5 built in class > stretched-link

Only thing is that the button to ADD a category was displayed below the stretched link, so never clickable: i removed that altogether as that is something I do not think users will use a lot from the cpanel.
 
#### Testing Instructions
install PR, watch and behold... cpanel cards are now clickable :)


rich20:
- [x] Success :+1: Now I can close this: https://github.com/Kunena/Kunena-Forum/issues/8945

PR #9568, opened 2023-10-18
[6.1] Wrong http error code thrown when profile page may not be read by guests

c-schmitz:
#### Summary of Changes 

This PR fixes two issues. When trying to access a profile page and guests are not allowed to view profile pages, you would see the correct login screen, but a http code 500 was thrown. This originated from the problem that $exception was thrown with keyword 'new' which reset the error code inside the exception.

After fixing that it would still throw a 404 (not found) instead of a 403 (no access).
Since  a $user would never exist in that scenario (guests don't have a user) I removed the condition and set it to 403 only.

#### Testing Instructions
Set in settings that guests may not see profile pages, then try to acess a profile page URL when not logged in. (copy the URL of a valid profile while being logged in). Check the network tab and see that  500 is thrown (before this patch is applied).
After applying the patch a proper 403 is thrown. Search engines like that much better.

### DIFF
--- a/src/libraries/kunena/src/User/KunenaUser.php
+++ b/src/libraries/kunena/src/User/KunenaUser.php
@@ -730,7 +730,7 @@ class KunenaUser extends CMSObject
         switch ($action) {
             case 'read':
                 if (!isset($this->registerDate) || (!$user->exists() && !$config->pubProfile)) {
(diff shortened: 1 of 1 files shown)


c-schmitz:
~~Great, will this patch be automatically also in 6.2?~~
Sorry, just saw you already did it ;)

xillibit:
In the future just do your PR against K6.2 branch because the K6.1 won't have any futher developpement

PR #9594, opened 2023-12-10
Fix: element (form) ID's must be unique on a page

Ruud68:
Pull Request for Potential Issue 
 
#### Summary of Changes 
elements on a page should have a unique ID, currently the quick reply forms all have the same form ID. This can cause errors / conflicts when trying to do CSS / JS functions on these forms
 
#### Testing Instructions
install PR, test if everything is still working.
Take special attantion when you have JS or CSS using the form ID "postform" as this ID now has been made unique with the message ID the postform is linked to.

### DIFF
--- a/src/site/template/aurelia/layouts/message/edit/quickreply.php
+++ b/src/site/template/aurelia/layouts/message/edit/quickreply.php
@@ -62,10 +62,10 @@ if ($me->canDoCaptcha() && KunenaConfig::getInstance()->quickReply) {
 }
 ?>
 
-    <?php if ($quick == 1) :
-        ?>
+<?php if ($quick == 1) :
+?>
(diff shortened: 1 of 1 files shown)


rich20:
I still don't recognise any difference (tested with CAPTCHA - Invisible reCAPTCHA and Captcha - reCAPTCHA v3 from JED). I still have to deactivate the first captcha if I want the second one to be displayed.
However, I get error messages (which have nothing to do with this fix), so I can only recognise which of the two captcha Kunena wants to display by the different error messages.

Ruud68:
Hi, this is not captcha related, that change I still need to do. First need to setup 6.3 Environment

rich20:
> Hi, this is not captcha related, that change I still need to do. First need to setup 6.3 Environment

Thanks, yes I had noticed, that's why I labelled my post "off topic".

PR #9626, opened 2024-02-22
Change: refactored config back-end view into J5

Ruud68:
Pull Request for Issue # . 
 
#### Summary of Changes 
refactored config view:

- new tabs
- new modal
- re introduced color when setting changed

#### Testing Instructions
the usual: try everything and see if it still works

### DIFF
--- a/src/admin/media/css/theme.min.css
+++ b/src/admin/media/css/theme.min.css
@@ -2994,6 +2994,10 @@ a:active {
 	border-radius: 50%
 }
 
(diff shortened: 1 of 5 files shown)


xillibit:
Why there yellow part at random places ? 
image

Ruud68:
Hi, when you change a config setting to a value other then the default setting a class is added to that row. That was always the case, I only reintroduced the css with the yellow color. I think it was this way in k5 and below.

PR #9628, opened 2024-02-23
Change: refactored logs back-end view into J5

Ruud68:
Pull Request for Issue # . 
 
#### Summary of Changes 
So this is a bigger refactoring.
@xillibit can you tell me what the grouping is that is used? I do not have this in my own log tables so I cannot test this as i have no clue as to how that works.
Also I removed the usage of the KunenaLogFinder as IMO this is not needed anymore when using the ListModel s getListQuery() method
Copying the log data is now done inline instead of in a modal, just click the data and it will be copied to the clipboard

#### Testing Instructions
The usual, sorting, filtering, etc.


Ruud68:
@xillibit there are two more views left (aside from the individual (form) views: but that is whole different 'beast' to tame :) )  statistics and trashbin

Trashbin I have data for (or can create data for): so this i will be able to refactor into J5 compliant code.
Statistics I have never seen / used myself: on all my sites this is empty so no clue as to how to get data into that view: without data it is hard to refactor. Do you have a clue as to how to get data into that statistics view?

xillibit:
> @xillibit can you tell me what the grouping is that is used? I do not have this in my own log tables so I cannot test this as i have no clue as to how that works.

What are you calling grouping ?

Ruud68:
the $this->group as it is used here: https://github.com/Kunena/Kunena-Forum/blob/7bcd58f9fa7123b65157af4a5dfd0b2bd59022b3/src/admin/tmpl/logs/default.php#L256-L300

so the items can be grouped, but I have no idea how / where to do that. If it is not used, I can just delete it from the code.

xillibit:
For the group items it's not me which had wrote  this part of code and i don't remember the purpose of this

> Statistics I have never seen / used myself: on all my sites this is empty so no clue as to how to get data into that view: without data it is hard to refactor. Do you have a clue as to how to get data into that statistics view?

I have data on statistics on a site, i will send it on your mail later today

Ruud68:
The group is an array with all the columns in it, so group_type, group_user, group_category, etc.
I see that it is read via the LogsModel in populateState
```
        if ($this->getUserStateFromRequest($this->context . '.group.type', 'group_type', false, 'bool')) {
            $group['type'] = 'a.type';
        }
```
but there is nothing in the 6.2 / 6.3 branch code that sets these.

So I think that this is from an older version of Kunena where the filter / toggle that sets the grouping didn't make it into 6.1 / 6.2 / 6.3. (I even checked 5.2 and could not find it in there either)

My guess is that it is safe to remove, maybe @810 knows?

rich20:
This log only records actions by admins and moderators, but only if "Define if you want to log action or moderation = Yes" in the configuration. Nothing is recorded by other user groups.

**Log file**
.
Bildschirmfoto vom 2024-02-26 18-40-19

**Statistic**
.
Bildschirmfoto vom 2024-02-23 16-02-30

Ruud68:
@rich20 the log screenshot is not with the new.log backend view. It should have the j5 filterbar on top instead of the one showing.
Did you install the PR?
Also, nothing changed in the logging it self only in the logs view displaying the log entries

rich20:
> Did you install the PR?

Sorry, it was a screen without your PR. The logs are fine with your PR now.
.
Bildschirmfoto vom 2024-02-27 00-54-27

Ruud68:
Ok @rich20 thanks for testing and confirming that the PR works correct.
Do you also know anything (from history) about the possibility to group the entries in the view?
I found code to do that but there is no way in the interface (that i could find) to actually do it: so not sure if this code is a leftover from an old version that is not implemented in the newer versions

rich20:
There was never a possibility for grouping, as nothing is recorded from normal registered users. Only mod actions are recorded and apparently admins also belong to the moderator group.
But actually there is no moderator group, indeed moderators have extended rights, but neither Joomla nor Kunena have a moderator group by default. Apparently, users are assigned to this back end log group based on their access rights.

Ruud68:
What the code I found does (when configured) is group the log entries by e.g. type or operation or user or category etc.
Is that something you know how to configure / use @rich20 ?

rich20:
This is the basic setting and is only used to select what is displayed in the log. This setting is retained even after logging out.
In this screenshot (moderator = user richi) you can see the moderator actions that richi has carried out at the user Horst.
.
Bildschirmfoto vom 2024-02-27 14-14-20

Ruud68:
Yes, that is the filter (target_user = richi, but that is not what the group in de code does: that will create and display a group where the individual log entries are gone and the group + a total of the number of entries is displayed.

rich20:
When a moderator is selected on the right side in the user filter, then all actions of this moderator are displayed and all additional filters such as Select Category, Select by Time etc... work correctly. Except for the Operations filter, which never shows anything.
If you select a user in the left user filter, something is only displayed if a moderator has performed an action on this user. This is also correct, as only moderator operations are logged.
If a moderator is entered in the left user filter, an incomplete list is created, because types such as MOD or ERR are not displayed.

Ruud68:
I do not understand from your comment if there is an issue or not.
when selecting a moderator (left / first user filter), the operations filter works
image

rich20:
I'm sorry, I meant the "Select Operation" option.

xillibit:
For the logs, plugins the code was copied mostly from Joomla! views it could have been not finished

> Statistics I have never seen / used myself: on all my sites this is empty so no clue as to how to get data into that view: without data it is hard to refactor. Do you have a clue as to how to get data into that statistics view?

> > I have data on statistics on a site, i will send it on your mail later today

The model of stats use the KunenaLogFinder() to get data is a bit weird, i will try to export all tables to see

Ruud68:
Hi @xillibit I think I have figured out the statistics, working on it now.
Will push a PR one if these days ao you can test.
What I see is that it uses the data from the logs table. I have never used that view myself and I do not know what a use case for this view is: it just displays per user e.g. how many thanks, deletes, etc. But when you purge the log table then these values are 0...

What is that view used for?

rich20:
> Statistics I have never seen / used myself: on all my sites this is empty so no clue as to how to get data into that view

@xillibit  Configuration -> tab Users -> Define if you want to log action or moderation = Yes
You will then receive an entry in the log for each moderator or admin event. I have never seen an option for group selection or grouping. As it is present in version of Kunena 6.2. as it has always been. It was never more.

> What I see is that it uses the data from the logs table. I have never used that view myself and I do not know what a use case for this view is: it just displays per user e.g. how many thanks, deletes, etc. But when you purge the log table then these values are 0...

@Ruud68 This statistic only counts the number of thank yous, deleted posts, etc... by the moderators. This has never worked properly.

> But when you purge the log table then these values are 0...

The next time you act as a moderator (e.g. reply to a topic), the statistic is back but also many empty pages.
.

PR #9637, opened 2024-02-29
Refactor log clean view into modal making log controller / model / vi…

Ruud68:
…ew obsolete

Pull Request for Issue # . 
 
#### Summary of Changes 
The Log model / controller / view / tmpl was only there for the logs clean function.
I have moved that clean tmpl to a modal in the logs view.
I have moved the clean function from the log controller to the logs controller.

Now the LogModel, LogController, Log view and Log Tmpl that where basically copies of the logs equivalents can be removed.

This PR removes the obsolete files from the repositorie bu NOT from an installation when updating. @xillibit when this PR is accepted then the removed files should also be cleaned up from the site where the update is installed. Can you do that in the script as i have no knowledge of that?
 
#### Testing Instructions
clean your logs in the log view and see if it is actually cleaned


xillibit:
> This PR removes the obsolete files from the repositorie bu NOT from an installation when updating. @xillibit when this PR is accepted then the removed files should also be cleaned up from the site where the update is installed. Can you do that in the script as i have no knowledge of that?

yes i will do it after i have tested and merged this PR

Ruud68:
@xillibit cool, please note that in order to test correctly you should already remove the files from the test environment (as updating the package with this PR doesn't do that)

Ruud68:
PS as for the deleting of the files, maybe (just an idea) it is best to have a script preflight routine that will remove the:

administrator/components/com_kunena:

- api
- forms
- services
- tmpl
- src

Directories (and something similar for the kunena library directory and site component com_kunena directories)

This because I have also other 'left-over' / removed files on my production server from previous upates where not all removed files where also removed from the server it self.

Ruud68:
yeay! just did https://github.com/Kunena/Kunena-Forum/pull/9639 for the cleanup, had some time while waiting for a customer to give feedback :)

PR #9639, opened 2024-02-29
cleanup folders in install script

Ruud68:
Pull Request for Issue # . 
 
#### Summary of Changes 
added preflight folder cleanup. it is safe to remove folders in preflight as these will be then filled with the update or installation of the extension.
ASSUMPTION (!) here is that the site owner did not do any modifications / added files to the folders that are being deleted

It checks for the installed version and when that is above 6.0.0 and below 6.3.0 then the folders are handled.

#### Testing Instructions

build package with this PR in it and install it, there should not be any errors and everything should be working as it should.
All Kunena provided folders should now be exactly like in the install package without any left over files from previous versions that where not removed.


xillibit:
You should exclude the following files because the user could have added things in it : 

\src\site\template\aurelia\assets\scss\custom.scss
\src\site\template\aurelia\config\params.ini
\src\media\kunena\core\js\customckeditor_config.js (the user can personnalize the name of this file in template config it should start with custom)

Ruud68:
Removed the aurelia directory,
what is the installation path of the customckeditor_config.js file? As I cannot find this on my dev station

xillibit:
It's \joomla-51-dev-kunena-63-dev\media\kunena\core\js\

Ruud68:
Ok, this PR doesn't remove anything in the [site]/media folder

Issue #9646, opened 2024-03-05
Backend Log | Clean entries not possible


rich20:
**To Reproduce**
Click on the "Clean entries" button, select a short time period and click on Submit

**Result**
No entries are removed

**System information**
- Joomla version: ‎5.0.4-dev
- Joomla template: CASSIOPEIA
- Kunena version: 6.3.0-BETA2
- Kunena template: Aurelia
- Php version: 8.2.16
- Database version: 10.5.23-MariaDB

**Desktop**
- OS: Ubuntu 22.04.1 LTS
- Browser: Chromium, Firefox and Opera

Ruud68:
Hi @rich20 you explicitly say 'short period', does it work when you set a long period?
When e.g. setting 30, it would remove all log entries older then 30 days.
What short period are you trying with?

rich20:
Hi @Ruud68 
The default is 30 days, so only the logs from the last 30 days should be retained. I have tried it with 30 days, 10 days, 1 day, 300 days, 0 days, the result is always the same. All logs are retained, nothing is removed.

Ruud68:
Hi @rich20 
found the issue (on multiple modals).
refactored the modals to J5 but that upped the minimum Joomla version also to 5, so yesterday did a quick implementation to refactor these modals to J4.4 compatible.
Was a rush job to get the beta2 out off the door.

modals where working (visual), only the submitting of these modals not (was missing some additional onclick code.

fixed here: https://github.com/Kunena/Kunena-Forum/pull/9647

rich20:
Thanks, it works now.

PR #9647, opened 2024-03-06
Fix: refactored modals not working correct

Ruud68:
Pull Request for Issue # https://github.com/Kunena/Kunena-Forum/issues/9646. 
 
#### Summary of Changes 
the following refactored to 4.4 modals where not sumbitting data / form correct:

- logs > clean
- users > moderators
- users - subscribecategories
- categories > batchmove

#### Testing Instructions
test not only if modals work (they actually show), but if submitting / processing also works

### DIFF
--- a/src/admin/src/Controller/CategoriesController.php
+++ b/src/admin/src/Controller/CategoriesController.php
@@ -754,8 +754,9 @@ class CategoriesController extends KunenaController
      * @since   Kunena 5.1.0
      * @throws \Exception
      */
-    public function batchCategories(): bool
+    public function batchcategories(): bool
     {
+        // Function renamed from batchCategories in Kunena 6.3.0-BETA3 as this name collides with view / model
         if (!Session::checkToken('post')) {
             $this->app->enqueueMessage(Text::_('COM_KUNENA_ERROR_TOKEN'), 'error');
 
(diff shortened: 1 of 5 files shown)


xillibit:
> * users > moderators
> 
> * users - subscribecategories

Tested on Joomla! 4.4, when submit the modal it does nothing

> * categories > batchmove

image

Ruud68:
@xillibit let me check that tomorrow

Ruud68:
Ok, I'm sure I tested all (except the category batch one): go figure.
Same issue as with some other controller method names: batchCategories method is the same as var in view / model. Somehow PHP / Joomla doesn't like this any more.
So renamed the method to all lowercase and now it works.

presuming this was the only modal not working, this should now be okay.

Updated this PR with new commit

Ruud68:
Ok, just did a test and now the controller method is called okay. But now it is saying that I forgot to select a target category.

Turnout that the controller is getting the selected category from a field on the form that doesn't exist (and also didn't exist in K6.2 so this never worked?).
Anyway, did a last small change (in this PR) and should be working now.
Can you test @xillibit ?

xillibit:
There was two options in the categories batch one have never worked, so i have removed part in the template

PR #9651, opened 2024-03-11
Proposal: replace Joomla user filters with custom user filter in logs view

Ruud68:
Pull Request for Issue https://www.kunena.org/forum/k-6-3-0-support/168364-help-needed-logs-backend-view?start=0#230977 . 
 
#### Summary of Changes 
This PR replaces the two user filters (users / target users) in the logs view.
Currently the filters are based on the Joomla core users field, this PR changes these fields to a customized dropdown filter where only users are listed that have an entry in the logs table.
 
#### Testing Instructions

This is a proposal, I like the Joomla core users field, but the added value here is very limited and it doesn't allow us to overwrite the label (select user > select target user)

Let me know what you think so @xillibit can either decline or merge (after testing)

cc @rich20

### DIFF
--- a/src/admin/forms/filter_logs.xml
+++ b/src/admin/forms/filter_logs.xml
@@ -42,11 +42,19 @@
 			class="js-select-submit-on-change" />
 		<field
 			name="user"
-			type="user"
-			label="COM_KUNENA_LOG_USER_SORT_LABEL"
-			hint="COM_KUNENA_FILTER_SELECT_USER"
-			class="js-select-submit-on-change"
-			default="" />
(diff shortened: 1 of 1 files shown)


rich20:
For me, the user filter is fine as it is. But I have found problem, so I will create a new ticket for it.

Ruud68:
My preference is having the new filters as using the Joomla User fields can be confusing because they in label text do not differentiate between user and target user, further more the popup modal with all the filtering on the users is imo 'overkill'

rich20:
- [x] Succes

Issue #9654, opened 2024-03-13
The time zone selected in Joomla is not adopted by Kunena in the backend


rich20:
**Describe the bug**
https://www.kunena.org/forum/k-6-3-0-support/168364-help-needed-logs-backend-view?start=20#230997
Confirmed

**To Reproduce**
Select a time zone in Joomla, e.g. Amsterdam or another

**Result**
The time is only adjusted in the frontend, not in the Kunena log files. This leads to a time difference between the actions in the frontend and in the log file

Ruud68:
@rich, did a PR #9656 for this, can you check?

rich20:
Fixed by this PR https://github.com/Kunena/Kunena-Forum/pull/9656

PR #9656, opened 2024-03-14
Log date format

Ruud68:
Pull Request for Issue #9654 . 
 
#### Summary of Changes 
changed displaying of date to take into consideration the configured server setting for the timezone

#### Testing Instructions
in back-end view logs, check if the log entry date is using the timezone setting from the Joomla global config (date / time should be the same as date / time in front-end


rich20:
- [x] Success

Issue #9655, opened 2024-03-13
Error when trying to access the trash via the dashboard


rich20:
**Describe the bug**
https://www.kunena.org/forum/k-6-3-0-support/168371-kunena-6-3-0-beta3-dashboard-trash-error#230999
Confirmed

**To Reproduce**
Go to the Dashboard and click on Trash

**Actual result**
View not found [name, type, prefix]: trash, html, Administrator 
# | Function | Location
-- | -- | --
1 | () | JROOT/libraries/src/MVC/Controller/BaseController.php:881
2 | Joomla\CMS\MVC\Controller\BaseController->getView() | JROOT/libraries/src/MVC/Controller/BaseController.php:655
3 | Joomla\CMS\MVC\Controller\BaseController->display() | JROOT/administrator/components/com_kunena/src/Controller/DisplayController.php:62
4 | Kunena\Forum\Administrator\Controller\DisplayController->display() | JROOT/libraries/src/MVC/Controller/BaseController.php:730
5 | Joomla\CMS\MVC\Controller\BaseController->execute() | JROOT/libraries/src/Dispatcher/ComponentDispatcher.php:143
6 | Joomla\CMS\Dispatcher\ComponentDispatcher->dispatch() | JROOT/libraries/src/Component/ComponentHelper.php:361
7 | Joomla\CMS\Component\ComponentHelper::renderComponent() | JROOT/libraries/src/Application/AdministratorApplication.php:150
8 | Joomla\CMS\Application\AdministratorApplication->dispatch() | JROOT/libraries/src/Application/AdministratorApplication.php:195
9 | Joomla\CMS\Application\AdministratorApplication->doExecute() | JROOT/libraries/src/Application/CMSApplication.php:306
10 | Joomla\CMS\Application\CMSApplication->execute() | JROOT/administrator/includes/app.php:58
11 | require_once() | JROOT/administrator/index.php:32

**System information**
- Joomla version: ‎5.0.4-dev
- Joomla template: CASSIOPEIA
- Kunena version: 6.3.0-BET3-DEV
- Kunena template: Aurelia
- Php version: 8.2.16
- Database version: 10.5.23-MariaDB

**Desktop**
- OS: Ubuntu 22.04.1 LTS
- Browser: Chromium, Firefox and Opera

Ruud68:
Hi @rich20 , should be fixed here #9657, can you test?

rich20:
[...]

PR #9657, opened 2024-03-14
Fix: dashboard trash button didn't work anymore

Ruud68:
Pull Request for Issue #9655 . 
 
#### Summary of Changes
dashboard ison for trash was still pointing to old (removed) trash view
 
#### Testing Instructions
implement PR and click on the dasboard trash button, the trashs view should show without errors

### DIFF
--- a/src/admin/tmpl/cpanel/default.php
+++ b/src/admin/tmpl/cpanel/default.php
@@ -431,7 +431,7 @@ $count = KunenaStatistics::getInstance()->loadCategoryCount();
                             <div class="row align-items-center">
                                 <div class="col">
                                     <h6 class="mb-25">
-                                        <a href="<?php echo Route::_('index.php?option=com_kunena&view=trash'); ?>" class="stretched-link">
+                                        <a href="<?php echo Route::_('index.php?option=com_kunena&view=trashs'); ?>" class="stretched-link">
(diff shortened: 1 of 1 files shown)


rich20:
I still get the same error message

Ruud68:
The only thing this PR does is add an _s_ to the link, so first it was `..&view=trash` and now it is `..&view=trashs`

Can you check on if the link in your test site is correct, if not then I think the PR is not applied correct as I have tested this and for it now works correct

rich20:
The "s" at the end is still missing. But I have noticed if I delete the complete file in the folder cpanel, it is no longer in the folder also after a new installation. This will not installed.

Ruud68:
I just build an installation package with my PR branche.
Deleted the tml/cpanel directory (confirmed that the dashborad wasn't working due to mussing files) and installed the package: Installed without errors and dashboard with trash**s** is working correct
So guess there must be another issue on your server?

rich20:
Yes, now it works for me too, after I emptied the cpanel folder on my server installation. Then I installed your PR again. 
I can't say what exactly the reason was, but I only realised after some time that the cpanel default.php was never updated during the installation.
Now the cpanel folder is empty and it still works.
Can be merged

Issue #9649, opened 2024-03-06
Backend Category filter


rich20:
<strike>- [x] **Select Category** - all selection options work
- [x] **Select Category Level** - all selection options work
- [x] **Select Enabled = Published** - works
- [ ] **Select Enabled = Unpublished** - does not work
- [ ] **Select Access** - does not work at all
- [x] **Select Locked = No** - works
- [ ] **Select Locked = Yes** -  does not work
- [x] **Select Review = No** - works
- [ ] **Select Review = Yes** - does not work
- [x] **Select Polls = No** - works
- [ ] **Select Polls = Yes** - does not work
- [x] **Select Anonymous = No** - works  
- [ ] **Select Anonymous = Yes** - does not work
- [x] **Select Anonymous - Incorrect selection text.** instead of Anonymous user, the word must be **Yes**
and instead of Registered user the word **No** must be available for selection. This option only hides or shows categories in which anonymity is permitted.</strike>

Ruud68:
Hi, so I think this needs some more thought / clearification.

e.g. the enabled filter > unpublished shows all categories that are unpublished but only (!) when also the section they reside in are unpublished.

When refactoring the backend views only the interface has changed and not the actual filtering, so this (not tested as this for me never worked) is the same behavior as in 6.2 (where it probably didn't filter at all)

So maybe @xillibit can shed some light on this?

As for the access level filtering: correct that was on my list to fix (but that list got lost). Will do that together with the renaming of the anonymous seltion text (yes/no)

rich20:
> When refactoring the backend views only the interface has changed and not the actual filtering, so this (not tested as this for me never worked) is the same behavior as in 6.2 (where it probably didn't filter at all)

You are right, the filters did never work in the Kunena 6 series.

Ruud68:
So when getting the published status of a category it always takes into account the status of the parent. So if the parent is not disabled, then the category is also not disabled and without knowing the implications / the why that is, I am reluctant to change that: there must be a reason why it works this way... That is what we need to find out.

Ruud68:
@rich20 I am a bit lost on this as there where discussions in several places on this topic. Can you update the checklist and tell me what is in your opinion still not working correct?

rich20:
The first list is now out of date and I have crossed it out.
All filters work so far, but you also have to select the parent section or category. However, this should only be an additional feature.
[...]

PR #9659, opened 2024-03-15
Change: let category filters act on stored category data

Ruud68:
Pull Request for Issue https://github.com/Kunena/Kunena-Forum/issues/9649.

### Summary of Changes

This is a small change but it impacts all filtered results in the back-end categories filters.
e.g. when selecting unpublished it will display all unpublished categories regardless of the published status of the parent of the categories (section).
Same for review, polls, locked, anonymous
### Testing Instructions

implement PR and test all back-end category view filters and combinations of these filters


rich20:
- [x]  Now it's perfect

Ruud68:
Cool, it should only effect the back-end view, but also keep an eye out for any changes on the front-end where categories are involved

Issue #9670, opened 2024-04-17
{K6.2 / K6.3] Plugin Modals buttons not working on Joomla 5.1


Ruud68:
**Describe the bug**
Joomla 5.1 has an (undocumented) B/C break when it comes to modals.
This results in the plugin modals (kunena backend) not working correct anymore: the buttons save, save & close and close do not work any more, further more there are also buttons displayed on top which work but have an unexpected side effect (e.g. when clicking close, you would expect the modal to close but it doesn't)

Reported this here: https://github.com/joomla/joomla-cms/issues/43289

I have a 'quick' workaround, but pending the result of how Joomla is going to fix this bug it is better to wait to get some clearity on their approach.

I will do a PR here for K6.2 and 6.3 once we have an approach

xillibit:
I will make the stable release of K6.3 in a couple of days so it's useless to continue to maintain K6.2 now

Ruud68:
ok, that will save some time. K6.3 should still support J4 and J5.0 correct?

xillibit:
yes K6.3.x support both J!4.4.x and J!5.x

Ruud68:
ok, pr is in: #9672

PR #9672, opened 2024-04-18
Fix: refactor plugin modals to dialogs to workaround J5.1 bug

Ruud68:
Pull Request for Issue #9670  . 
 
#### Summary of Changes 
added new J5.1 dialogs when Joomla version is 5.1
 
#### Testing Instructions
test on both J5.1 and < J5.1

1. in back-end goto Kunena > plugins
2. click a plugin name

on J5.1 a dialog modal should open (save / save& close / close buttons on top
on  < J5.1 a BS modal should open (buttons in the footer)

3. test if all buttons work.

The J5.1 plugin dialog modal is missing the Toggle inline help button (the BS modal has that button in the footer).
I have crated a PR for that in the Joomla repository: appreciate if you could also test that and if okay report back as positive test result on the Joomla repository: https://github.com/joomla/joomla-cms/pull/43296

### DIFF
--- a/src/admin/tmpl/plugins/default.php
+++ b/src/admin/tmpl/plugins/default.php
@@ -15,17 +15,19 @@ use Joomla\CMS\HTML\HTMLHelper;
 use Joomla\CMS\Language\Text;
 use Joomla\CMS\Layout\LayoutHelper;
 use Joomla\CMS\Router\Route;
+use Joomla\CMS\Version;
 use Kunena\Forum\Libraries\Version\KunenaVersion;
 
 /** @var \Joomla\CMS\WebAsset\WebAssetManager $wa */
 $wa = $this->document->getWebAssetManager();
 $wa->useScript('table.columns')
     ->useScript('multiselect');
-$app       = Factory::getApplication();
-$user      = $this->getCurrentUser();
-$userId    = $user->id;
-$listOrder = $this->escape($this->state->get('list.ordering'));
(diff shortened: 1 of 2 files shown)


xillibit:
It-is need to apply the PR4326 from Joomla! to test it ?

Ruud68:
> It-is need to apply the PR4326 from Joomla! to test it ?

No, not needed.
It should work without the Joomla PR.

The Joomla PR adds the button 'Toggle inline help' that is missing in the new 5.1 dialog modal.
You can test that separately if you have time and energy to do so :)

src/admin/tmpl/plugins/default.php
```diff
@@ -64,21 +66,39 @@
...
+                                $linkModal = HTMLHelper::_(
+                                    'link',
+                                    '#plugin' . $item->extension_id . 'Modal',
+                                    $item->name,
+                                    'data-bs-toggle="modal" id="title-' . $item->extension_id . '"'
+                                );
+                            } else {
+                                $popupOptions = [
+                                    'popupType'  => 'iframe',
+                                    'textHeader' => Text::sprintf('COM_OCHSUBSCRIPTIONS_CPANEL_MANAGER_PLUGIN', $item->name),
```
xillibit: you have introduced a language key string from OCHSUBSRIPTIONS

Ruud68: my bad, fixed!

PR #9685, opened 2024-05-04
Wrong time stamp shown when listing topics/messages in trashbin

c-schmitz:
#### Summary of Changes 
When listing messages or topics in trashbin, it would show bogus timestamps because the UNIX timestamp was not converted properly. 

#### Testing Instructions
Delete  message to trashbin, Check out the original creation time, compare with time shown in trashbin.
Before this patch it was different, the message in the trashbin would show a timestamp on the same day, but a different time.
After applying the patch it shows correctly.


Ruud68:
Tested and works okay, good catch @c-schmitz !

c-schmitz:
Yeah. I tried to trace some deleted spam messages and was wondering about the timestamps not matching the web logs.
:-)

PR #9765, opened 2024-11-13
Change: add title to labels on topic moderation icons

Ruud68:
Pull Request for Issue # . 
 
#### Summary of Changes 
When moderating a topic, I always struggle setting the correct topic status as there are only icons (which in my case can be the same).
This change will add the 'icon title' as a HTML title attribute to the icon label. This way, hovering the icons will show the title in the browser.

image

 
#### Testing Instructions
This needs to be tested with different kind op topic labels and topic icons (svg, fa, etc.)


rich20:
- [x] Success

Tested with same symbols and different labelling
.
g335939
Works with all icons, only Images Legacy are not displayed in this place (only the correct text) if Selectable Topic Icons is set to No.
But this has nothing to do with this fix, it was the same before.

Issue #9766, opened 2024-11-14
Private attachments are in the topic history


rich20:
**To Reproduce**
Write an answer and add a private attachment
Login with another user
Click on Reply and scroll down

**Actual result**
The attachment is available in the Topic History

**System information**
- Joomla version: ‎5.2.1-dev
- Joomla template: CASSIOPEIA
- Kunena version: 6.3.7-DEV
- Kunena template: Aurelia
- Php version: 8.2.24 & 8.3.12
- Database version: 10.5.26-MariaDB

**Desktop**
- OS: Ubuntu 22.04.5 LTS
- Browser: Chromium, Firefox and Opera

Pinkeltje:
Created Pull Request 
Still don't know how to remove old commits...........Sorry

Pinkeltje:
Messed up. New pull request made. #9768

PR #9768, opened 2024-11-16
Make whatsapp icon clickable and hide private attachments in edit history

Pinkeltje:
No issue added but post from forum
[https://www.kunena.org/forum/k-6-3-0-support/168605-whatsapp-icon-not-displaying-correctly](url)
And issue #9766


rich20:
<s>Private attachments are now generally no longer displayed in the history. But I don't think that's a problem. What do you think about this?</s>

rich20:
Private attachments are now generally no longer displayed in the history (not even for authorised persons). But I don't think that's a problem. What do you think about this?

Pinkeltje:
Don't think it's a problem as it is only in the history. As long as they are displayed in de message, for me it's ok.

Issue #9769, opened 2024-11-17
Edit mode : Protected attachments are always displayed as private


rich20:
**Describe the bug**
https://github.com/Kunena/Kunena-Forum/pull/9764
I found one more thing as I had tested protected attachments.  Protected attachments are always shown as private in edit mode. 

**To Reproduce**
1. set in the configuration Protect Attachments to Yes
2. upload some attachments (it does not matter whether it is inserted into the message or not)
3. Submit and then edit the message

**Result**
All attachments are displayed as private

**System information**
- Joomla version: ‎5.2.2-dev
- Joomla template: CASSIOPEIA
- Kunena version: 6.3.7-DEV
- Kunena template: Aurelia
- Php version: 8.3.12
- Database version: 10.11.8-MariaDB

**Desktop**
- OS: Ubuntu 22.04.5 LTS
- Browser: Chromium, Firefox and Opera

Pinkeltje:
Will check as soon as 6.4.0 B3 (or next version) is working.

PR #9787, opened 2024-12-21
Handle protected attachments the same as non-protected attachments

Pinkeltje:
Pull Request for Issue #9769  . 
 
#### Summary of Changes 
When setting Protect Attachments to Yes, protected attachments are now "behaving" like non-protected attachments
When testing discovered another bug: When saving message first and then edit and insert attachment, id was not retreived. Result: `[attachment=undefined]hash[/attachment]`
#### Testing Instructions
Please test carefully, because there ar so many variables in this file.

### DIFF
--- a/src/media/kunena/core/js/upload.main.js
+++ b/src/media/kunena/core/js/upload.main.js
@@ -1,158 +1,25 @@
 /**
- * Kunena Component
- * @package Kunena.Media
- *
- * @copyright     Copyright (C) 2008 - @currentyear@ Kunena Team. All rights reserved.
- * @license https://www.gnu.org/copyleft/gpl.html GNU/GPL
- * @link https://www.kunena.org
- **/
 
+ * Kunena Component
+
+ * @package Kunena.Media
+
+ *
+
+ * @copyright     Copyright (C) 2008 - @currentyear@ Kunena Team. All rights reserved.
+
+ * @license https://www.gnu.org/copyleft/gpl.html GNU/GPL
+
+ * @link https://www.kunena.org
+
+ **/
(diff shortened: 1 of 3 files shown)


rich20:
Now it doesn't work properly even without protecting attachments. I have uploaded three attachments. 
One not inserted in the message
one inserted in the message
set one as private
Bildschirmfoto vom 2024-12-21 17-35-07

--------------------------------------
## Result if you click on edit
g20397

Pinkeltje:
Think it is ok now. Took me a while to find that protected attachments have value 1 in field protected in attachments table where as private attachments have value 32.

Pinkeltje:
@xillibit latest commit is concerning 
https://www.kunena.org/forum/k-6-3-0-support/168674-incorrect-use-of-label-for-from_element#232590
Chrome thinks that labels should match id.

xillibit:
My commit change for the title input for quickreply on a subject, when i look with developper tools i have label and id matching : `<div class="form-group">
                                <label for="subject103" class="col-md-12 control-label" style="padding:0;">
                                    **Subject**:
                                </label>
                                <input type="text" id="subject103" name="subject" class="form-control" maxlength="80" value="topic test">
                            </div>`

Pinkeltje:
Yes, that one is ok. So shouldn't be changed. Maybe create seperate issue for Chrome and label "errors". My pull request was added automatically to this pull request, but it gets confusing now.

rich20:
- [x] Fix for protected attachments - Success
- [x] Label issue Chrome - Chromium no longer shows an error, so I think it also works with Crome

PR #9810, opened 2025-02-01
Update scss remove margin for birthdate and change layout for user profile fields on mobile

Pinkeltje:
#### Summary of Changes 
Birthdate field is not nicely aligned in edit profile

 
2025-02-01 Screenshot 833

User profile fields on mobile
#9815

### DIFF
--- a/src/site/template/aurelia/assets/scss/general.scss
+++ b/src/site/template/aurelia/assets/scss/general.scss
@@ -1089,7 +1089,7 @@ ul.inline > li > a > I {
 }
 
 #editprofile #birthdate {
-    margin-left: -15px;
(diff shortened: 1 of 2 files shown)


rich20:
- [x] Success

rich20:
It works as it should, but you can remove the prefixes #kunena.layout as these scss files are only called by Kunena anyway. It also works without. You only need the prefixes if you insert codes in other files such as user.css or Joomla template.css.

Pinkeltje:
Already did that.

rich20:
You were quicker to edit the pull, than I was written my previous post. :1st_place_medal: 

- [x] Success

Issue #9843, opened 2025-03-12
Category moderator becomes global moderator


rich20:
**To Reproduce**
1. In the backend, go to Kunena -> Users and select a user with the checkbox
2. Select Assign Moderator(s) at the top of the action menu
3. Select a category for which the moderator should be responsible

**Actual result**
He becomes a global moderator

**Screenshots**

![Image](https://github.com/user-attachments/assets/2b4a8def-faaa-4475-9731-6008c7f21910)

**System information**
- Joomla version: ‎5.2.6-dev
- Joomla template: CASSIOPEIA
- Kunena version: 6.3.10-DEV
- Kunena template: Aurelia
- Php version: 8.3.14
- Database version: 10.11.8-MariaDB

**Desktop**
- OS: Ubuntu 24.10
- Browser: Chromium, Firefox and Opera

Ruud68:
#9844 created for this, can you test @rich20 ?

PR #9844, opened 2025-03-12
Fix: Category moderator becomes global moderator

Ruud68:
Pull Request for Issue #9843 . 
 
#### Summary of Changes
removed automatically added 'preselect' value, and set 'selected' to none (emty array instead of 0 which is actually global moderator
 
#### Testing Instructions
Only tested this in the backend as reproduced by instructions in issue.
not sure if this is also impacting other parts of kunena (@xillibit )


xillibit:
You can't promote an user as moderator outside of backend

Ruud68:
Sorry, and yes, only changed the back-end users model here for the users view, so this change should only be used on that view / modal popup in that view.

rich20:
> You can't promote an user as moderator outside of backend

I don't know what this means, but for me this fix works as it should.

PR #9866, opened 2025-04-30
Change: Added filesize formatted unit converter method

Ruud68:
Pull Request for Issue ##9865 . 
 
#### Summary of Changes 
Current file size formatted unit converter was a simple divide by 1024 and round function with only KB numbers.

This PR creates a full blown file size formatted unit converter method that can be invoked using the htmlhelper. It has support for B, KB, MB, GB and TB and is using language strings.
Currently implemented to round to 2 decimals, but that can be changed via a parameter passed to this method (I personally think that 2 decimals is better especially in larger file sizes)
 
#### Testing Instructions
apply this PR and check if the file sizes are displayed correct, test with different sizes attachments, small to large.


rich20:
- [x] Success (but I haven't tested with GB or TB yet 😀 )

Ruud68:
Thanks @rich20  for testing, if the B, KB, MB work then the GB and TB also work :)))

@xillibit, as this code uses different language string, templates (other then aurelia) also need to be changed as they use the old language string that is not present anymore.

I ran into this with my own template. Now when a new release comes out, I always 'sync' my template: so for me this is not an issue. But  it will lead to displaying the old language string on templates that have not been updated but that will be using Kunena version with this PR.

One way around would be to keep the old (in this PR removed) language string.

image

xillibit:
That's for third parties template developper, Kunena brings only one template. I take care only to upload the files modified on crowdin for translators

Issue #9669, opened 2024-04-17
Undefined constant "Kunena\Forum\Libraries\Factory\KPATH_ADMIN"


pfitzer:
**Describe the bug**
Building the joomla index with the cli command ends with error: Undefined constant "Kunena\Forum\Libraries\Factory\KPATH_ADMIN"

**To Reproduce**
1. ssh connect to your server and cd to joomla installation dir
2. run `php cli/joomla.php finder:index`

**Expected behavior**
Command should run without any errors

**Actual result**
```
php cli/joomla.php finder:index
Finder Indexer
==========================

 Starting Indexer
 Setting up Smart Search plugins

In KunenaFactory.php line 215:

  Undefined constant "Kunena\Forum\Libraries\Factory\KPATH_ADMIN"
```

**System information**

Joomla version: 4.4.4
Kunena version: 6.2.6
Php version: 8.3.6

pfitzer:
I have tested this patch and there is another error showing know

```
 php cli/joomla.php finder:index
Finder Indexer
==========================

 Starting Indexer
 Setting up Smart Search plugins
 Setup 15135 items in 0.032 seconds.

In KunenaRoute.php line 902:

  Call to undefined method Joomla\CMS\Application\ConsoleApplication::getMenu()
  ```

GJSchaller:
Hello, I am running into this issue as well - is there any update to resolve this?

GJSchaller:
This may be related: https://github.com/joomla/joomla-cms/discussions/43391

Ruud68:
hi @GJSchaller and @pfitzer  , can you test #9869

GJSchaller:
@Ruud68 - that fixed it, thank you!

Ruud68:
@GJSchaller cool! also can you test / confirm that everything else is also working correct? Special look out for generated URL's  on front-end (Kunena) as that code changed.
Haven't tested that myself, just checked my self if error for cli was gone.

GJSchaller:
I'll give it a shakedown in a few hours - just started my work day here, I can take a look when work is over.

Ruud68:
no rush :)

xillibit:
By doing `php cli/joomla.php finder:index purge` i have this warning :

`Warning: Attempt to read property "id" on null in C:\Users\floflo\git\Kunena-Forum-6.4\src\libraries\kunena\src\User\KunenaUserHelper.php on line 112`

`Call Stack:`
[...]

PR #9869, opened 2025-05-06
Fixed: php cli/joomla.php finder:index

Ruud68:
Pull Request for Issue #9669 . 
 
#### Summary of Changes 
the ConsoleApplcations doesn't have all the classes that the Site- and AdministratorApplication.
Kunena only is aware of Site or Administrator (isClient) and as such doesn't handle e.g. updating the smart search index via the console correct.

This PR returns the SiteApplication when in the cli client in those methods that rely either on Site- or AdministratorApplcation being returned by FactorygetApplication().
 
#### Testing Instructions
run ```php cli/joomla.php finder:index``` from the terminal. There should be no errors.

When your site is on a subdirectory, the you also need to set the live_site url in the Joomla configuration: the CLI is NOT aware of the URL that the site is on as that information comes from $_SERVER variables which are not set in the console

### DIFF
--- a/src/libraries/kunena/src/Access/KunenaAccess.php
+++ b/src/libraries/kunena/src/Access/KunenaAccess.php
@@ -145,8 +145,9 @@ class KunenaAccess
         $this->moderatorsByUserid = [];
 
         // Reset read KunenaAccess for the current session
-        $me = KunenaUserHelper::getMyself();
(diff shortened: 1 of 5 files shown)


xillibit:
Thanks i had looked to this issue this last days but i didn't find how to fix it

PR #9871, opened 2025-05-08
Fix: Warning: Attempt to read property "id" on null when using cli

Ruud68:
Pull Request for Issue #9669 . 
 
#### Summary of Changes 
As suspected, cli doesn't know how to handle user methods, so when asking what user is logged in (getIdentity()) it will not result in a user object, but in null.
This PR will handle this situation and will set the ID for the user to 0 (guest)

#### Testing Instructions
turn on error reporting to maximum in Joomla global config
run php cli/joomla.php finder:index

there should be no errors.
Make sure that the indexing works as that is where the difference is.
When indexing is started from the backend (manual), the there is a user logged in. The user is used to determine category access. When run from the cli, there is no user logged in.

I do not think this impacts the search index, but as I do not use that functionality myself I have no means and no knowledge on how to test this.

### DIFF
--- a/src/libraries/kunena/src/User/KunenaUserHelper.php
+++ b/src/libraries/kunena/src/User/KunenaUserHelper.php
@@ -109,8 +109,9 @@ abstract class KunenaUserHelper
      */
     public static function initialize(): void
     {
-        $id        = Factory::getApplication()->getIdentity()->id;
-        self::$_me = self::$_instances [$id] = new KunenaUser($id);
+        // Get the user ide for the current user, note that when run from the cli we need to set it to 0 as cli cannot handle user methods like getIdentity()
+        $id        = Factory::getApplication()->getIdentity()->id ?? 0;
+        self::$_me = self::$_instances[$id] = new KunenaUser($id);
 
         // Initialize avatar if configured.
         $avatars = KunenaFactory::getAvatarIntegration();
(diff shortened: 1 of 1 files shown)


xillibit:
just tested it works

GJSchaller:
I'll see if I can test this when it comes to content that is restricted to non-guest users.  One of my Joomla / Kunena sites is a documentation site where specific categories are restricted to specific groups, and are not public.  If the items still get indexed, we should be good.

GJSchaller:
A brief test shows that the re-indexing (as seen from the Search component in Joomla's admin panel) did index everything, including restricted content, so this should be good to go.

Ruud68:
@GJSchaller so to make sure: the reindexing from backend has same index results as the indexes created via cli?

GJSchaller:
I didn't reindex from the back end, but I don't see a need to.

When I ran the indexing from the CLI, I ran it with --purge, to force a full rebuild of the index.

After the index ran from the CLI, I logged into the back end, and I can see content in the Smart Search component that is restricted to non-guest users.

Does that help?

Ruud68:
Can you from the back-end do a clear and then an index. Can you then confirm that the indexed result is the same as the cli index (with purge).
The difference is that with the cli we 'assume' guest while in the back-end it is the admin user who is doing the indexing (and as such the assigned rights for the categories are / might be different).

GJSchaller:
I followed the steps, and the result is the same.  Same number of index results, and the same items appear in the same order once the index is rebuilt.

Ruud68:
Top @GJSchaller thanks for testing and confirming!

PR #9887, opened 2025-06-18
Refactor Captcha handling

Ruud68:
Pull Request for Issue # (no issue yet) . 
 
#### Summary of Changes 
Currently we are triggering the captcha plugins directly by importing the plugin and then triggering events for onInit, onDisplay and onCheckAnswer

We trigger these events using the 'new' events methods: so triggering via the dispatcher which requires an Event parameter that holds the values.

So far so good.... but while refactoring my own ochCaptcha to be Joomla 6 compliant I see that the Joomla captcha implementation will (one day) move away from triggered events and should (must?) implement only one triggered event (onCaptchaSetup) that will setup the captcha as a so called provider (CaptchaProviderInterface). This provider will have functions: display, checkAnswer, setupField and as such it is moving away from the respective onDisplay, onCheckAnswer and onSetupField we had in 'old' captcha plugins.

This will get 'messy' sooner or later with captcha plugins that are refactored to implement the subscriber interface: this will then lead to errors for Kunena as by implementing the subscriberinterface, the Event used in the dispatcher->trigger is not 'converted' to a string, but passed as the actual Event > resulting in a hard error as the functions in the plugin require strings as parameter and not Events.

This PR lets Kunena handle all the possible captcha implementations by using the Joomla Core Captcha class.
We instantiate the captcha by using this class (Captcha::getInstance),  and we call the 'initialise' method (that will be agnostic of the captcha's implementation (so for new it will do a init, for old it will trigger the event onInit, etc.)

~So although currently nothing is broken,  this is a 'fix' for future issues especially as the Core captcha's are being replaced by commercial / free captcha's.~
Current Kunena implementation (without this PR) will not display the captcha when this is implemented in the new way.

#### Testing Instructions
implement PR
try creating a topic or a reply to a topic as guest where a captcha is required.
repeat using different captcha's


xillibit:
This kind of refactoring should be done against K6.5 not K6.4

Ruud68:
Hi Xillibit, didn't know that was a thing already :)
just did another PR #9888 against 6.5 branch

xillibit:
ok good, closed in favor of https://github.com/Kunena/Kunena-Forum/pull/9888

PR #9893, opened 2025-07-14
Refactored system plugin > J6 compatible

Ruud68:
Pull Request for Issue # . 
 
#### Summary of Changes 
 discussed here: https://www.kunena.org/forum/general-questions-and-how-tos/168900-joomla-6-support
 
#### Testing Instructions


xillibit:
For the cleanExpiredBans() i'am agree that he should moved to be used as cron job instead of loading on every page

I don't remember what was the purpose to have onKunenaGetConfiguration(), so yes i think it can be removed

Ruud68:
> For the cleanExpiredBans() i'am agree that he should moved to be used as cron job instead of loading on every page
> 
I will (probably after my holidays) create a task plugin that we can use for these kind of chores: there are probably more things that we can add to that plugin then.

> I don't remember what was the purpose to have onKunenaGetConfiguration(), so yes i think it can be removed

What it allows you to do (and probably what it originally was intended for) is to add a plugin's parameters part of the Kunena Config class: That way you can get plugin parameters using the kuenanconfig.
I myself do not currently see any use case for this and as said: it currently doesn't do anything other then fetching the plugin parameters and then 'nothing'.
So I think this is save to remove.

Maybe good to get some Kunena / template / 3rd party developers on board here so they can share their use cases as well. I only know what I use myself and don't think I'm using it to its full potential :)

Ruud68:
Ok so followed the code for onKunenaGetConfiguration.
that should (but it doesn't) add the plugin parameters that listens to this event to KunenaConfig->plugins

The only instance where this is used is in KunenaParser::prepareContent

There it does: $config       = KunenaFactory::getConfig()->getPlugin('plg_system_kunena');

But (and just checked) $config is always 'empty' as onKunenaGetConfig is not working.

This means that the whole KunenaParser::prepareContent is not working correct as it doesn't use the configured values.
In fact, it does nothing other then just return $content unmodified because it default the events to false (instead of the configured events array)

I'll make an extra PR to fix this so you can test.

Ruud68:
#9894

Ruud68:
> > For the cleanExpiredBans() i'am agree that he should moved to be used as cron job instead of loading on every page
> 
> I will (probably after my holidays) create a task plugin that we can use for these kind of chores: there are probably more things that we can add to that plugin then.
> 
#9897

PR #9897, opened 2025-07-16
New: Kunena task plugin

Ruud68:
Pull Request for Issue # . 
 
#### Summary of Changes 
created new Kunena Task Plugin,
currently has 1 task: remove Expired Bans

when installing plugin it is automatically enabled and the task is created to run every hour.
 
#### Testing Instructions

### DIFF
--- a/src/plugins/plg_task_kunena/kunena.script.php
+++ b/src/plugins/plg_task_kunena/kunena.script.php
@@ -1,0 +1,149 @@
+<?php
+
+/**
+ * Kunena Package
+ *
+ * @package        Kunena.Package
+ *
+ * @copyright      Copyright (C) 2008 - @currentyear@ Kunena Team. All rights reserved.
+ * @license        https://www.gnu.org/copyleft/gpl.html GNU/GPL
+ * @link           https://www.kunena.org
+ **/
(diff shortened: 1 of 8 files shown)


xillibit:
Good idea to use the scheduled tasks feature available in Joomla!

Ruud68:
This can function as a basis for more (future) tasks. Can't think of any myself though :)

note that when this gets merged, the system plugin needs to be stripped of that functionality #remindertoself

src/plugins/plg_task_kunena/kunena.script.php
```diff
@@ -0,0 +1,149 @@
...
+        $db->setQuery($query);
+        $result = $db->loadResult();
+
+        if ($result) {
+            // We are already setup
+            return;
+        }
+
+        // Get the plugin parameters
+        $params = new Registry();
```
xillibit, line 128: $params i not used, can-be removed ? Or something is missing ?

Ruud68: Can be removed
 Left over from copy from my own code. If the task is created and enables them it is working correcy

PR #9899, opened 2025-07-17
Implement KunenaGetAvatarEvent

Ruud68:
Pull Request for Issue # . 
Joomla for gave us 'Events' as classes. Starting with the SubscriberInterface on plugins, these plugins now require events as parameters.
So the Joomla 3 way was:
```public function onKunenaGetAvatar() {return result;}```
With the subscriber interface it becomes:
```
public function onKunenaGetAvatar(Event $event) {
...
$event->setArgument('result', ...the previous return value...);
}
```

Joomla core handles the old way (returning the value) so that that still works in the new way (with events).

In Joomla 5 Core started to implement concrete events, so not the generic Event, but e.g. ContentPrepareEvent
Again they build it in such a way that the old plugins that didn't use events, still worked.

What I see in the deprecation notices in Joomla 5 is that that time is over and we should be starting to implement events / concrete events. A lot of things will get removed from J6. Currently not sure what exactly as in Joomla 6 tey e.g. state that a function will be removed in J6 but it is still in there and J6 is also still actively using it.

 
#### Summary of Changes 
As a proof of concept I created a concrete event: KunenaGetAvatarEvent that is triggered with the onKunenaGetAvatar dispatch.
This event has basically two functions: setAvatar (for the plugin) and getAvatar for the function wanting the avatar class.

But as there might be plugins out there that we do not have 'control' over, the created concrete class will also work for these plugins, so it will work for:

- return new KunenaAvatar....($this->params);
- $event->setArgument('result', [new KunenaAvatar...($this->params)];
- $event->setAvatar(new KunenaAvatar....($this->params));

I think that is this way we handle all the potential use cases out there, where we can start migrating our plugins to subscriber plugins with concrete events.

#### Testing Instructions
play with the different avatar providing plugins and see if they still work.

#### Assumptions
I assumed that there will always be one plugin that will provide an avatar or no plugin. So when you have multiple plugins that provide avatars, it will only use the avatar provided by the first responding (ordered) plugin. (This is wat the old implementation also did)

#### Side note
I think we can remove the EasyProfile plugin: don't think they made it to J4

We should also check the other plugins if these are still a live and kicking


Ruud68:
Ok, so last thing before going on holidays: added also KunenagetProfileEvent and refactored the plg_kunena_kunena into a new subscriber plugin: this can then function as reference for the other plugins.
Same for the events created, there are still a lot left that need creating / converting.

I also set a deprecated on KunenaAvater getEditURL function as i cannot find any place in kunena not in addons where this is actually used?

xillibit:
> I think we can remove the EasyProfile plugin: don't think they made it to J4

Yes you are right it's not updated since 7 years : https://extensions.joomla.org/extension/easy-profile/, so this plugin can be removed

> I also set a deprecated on KunenaAvater getEditURL function as i cannot find any place in kunena not in addons where this is actually used?

Maybe i had added it and changed the way then to get the link to profile edit page

PR #9902, opened 2025-08-11
Move events from Admin to Library

Ruud68:
Pull Request for Issue # . 
 
#### Summary of Changes 
 Moved Events from administrator to Library

#### Testing Instructions


xillibit:
I had added the Event for private and activity but in the plugin kunena_kunena you had added two helpers : https://github.com/Kunena/Kunena-Forum/tree/K6.5/src/plugins/plg_kunena_kunena/src/Helper for avatar and profile. What is the purpose of that ? Can-you look if needed to add it for activity and private ?

Ruud68:
Hi @xillibit ,
the plg_kunena_kunena I have refactored into the new 'subscriber' plugin.
so this plugin now extends CMSPlugin and implements SubscriberInterface.
All other plugins should also be refactored into this layout.
in this layout you then have locations for the specific files in the plugin.
The main class will be in directory src/Extension
The other files can then be located into the src/Helper directory (as that is what these classes do: help the main class provide the correct information for the (in this case) event.

So when we refactor the other plugins, the Helper directory will need to be created for them and the files will need to be in there.

PR #9907, opened 2025-08-26
Localstorage fix + announcement close remember

Ruud68:
Pull Request for Issue # https://www.kunena.org/forum/general-questions-and-how-tos/168920-marking-kunena-announcement-as-read. 
 
#### Summary of Changes 
Te site template has a value that, according to it's description, does the following: Load Localstorage > Remember the collapse state of the windows. Like announcements

due to changes in the aurelia template using Bootstrap collapse, the localstorage.js script did't work anymore. The script also didn't work on announcements as these are currently implemented as Bootrsap Alerts.

Furthermore not all collapse contaners had the correct initial classes (collapse show) which lead to the behavior that a first click on (e.g. stats) would try to open the already open stats widget. Same for who's online and write access

Note: I have create a vanilla JS script instead of a jQuery one (as I think this is the way forward, and my jQuery is not fluent :))
 
#### Testing Instructions
in Aurelia > features > turn local Storage on (yes)

1. try to toggle a collapse on e.g. the index page, do a page refresh (f5) and check if the state is remembered: so when you closed e.g. the statistics, they should be closed on a page reload.

2. create an alert and close this alert. on page reload the alert should not show.


Ruud68:
Added a way to clear the localstorage: in the kunena page where the script is loaded, add ?clear-kunena-localstorage=true

In the browser console you should see the number of removed storage items (if any), and all clollapses should be shown again as well as the announcement(s)

rich20:
- [x] It works for announcements
- [x] It works now on the index page when you collapse a section 
- [ ] It just doesn't work for category descriptions and category headings (but I don't know if that's even planned for those options).

> Added a way to clear the localstorage: in the kunena page where the script is loaded, add ?clear-kunena-localstorage=true

> In the browser console you should see the number of removed storage items (if any), and all clollapses should be shown again as well as the announcement(s)

In Profile -> Edit -> Forum Settings -> Clear Cache, you can also undo everything.

Ruud68:
Hi @rich20 can you test again, i have refactored the JS a bit and changed two layouts (for category hearder desriptions).

> In Profile -> Edit -> Forum Settings -> Clear Cache, you can also undo everything.

Didn't know that as I do not use the kunena profile, but I just checked it and this is actually considered a bug by me: it clears ALL the local storage, so also for e.g. content editor settings or back-end saved tab states etc.
I have refactored this button to now only clear the kunena stored settings and leave the other settings as is. (I changed the user > setting > layout . default.php file for this.

rich20:
-  [x] all success

PR #9909, opened 2025-09-09
Refactor: plg_kunena_community

Ruud68:
Pull Request for Issue # . 
 
#### Summary of Changes 

- Refactored the plugin kunena community, this is untested as I do not have the community extension.
- Refactored the plugin kunena comprofiler, also untested as i do not have the CB extension
- Refactored the plugin kunena easyblog also untested as i do not have the Easyblog extension
- Refactored the plugin kunena gravatar, also untested as i do not have the a gravatar
- Refactored the plugin kunena uddeim, also untested as i do not have uddeim component
- Refactored privacy plugin into subscriber plugin
- Refactored quickicon kunena plugin into service provider / subscriber plugin
- Refactored sampledata kunena plugin into service provider / subscriber plugin
- 
please test and report if things are working and if not, please give detailed information regarding what not

#### Testing Instructions


Ruud68:
note for easylog and uddeim, as they use the respective components libraries, the name spacing for instantiating new classes for these components should be validated / checked / corrected

Ruud68:
Done for now :)

Todo:

- plg_finder_kunena
- plg_kunena_easyblog
- plg_kunena_finder

Ruud68:
Tested:

- installation on existing site works
- new installation on Joomla 6 works
- sample data install works
- quickicon works
- gravatar works (quick test)

Plugins Todo:

- plg_finder_kunena
- plg_kunena_easyblog
- plg_kunena_finder

Let's first test if the other plugins are working correct.

xillibit:
- With Jomsocial integration : 

Namespace declaration missing in : Community.php
image

- With Community Builder integration : 
image

- With Uddeim integration : see my comment

src/plugins/plg_kunena_uddeim/src/Extension/Uddeim.php
```diff
@@ -0,0 +1,120 @@
...
+use Joomla\Event\SubscriberInterface;
+use Kunena\Forum\Libraries\Event\KunenaGetPrivateEvent;
+use Kunena\Forum\Libraries\Forum\KunenaForum;
+use Kunena\Forum\Plugin\Kunena\Uddeim\KunenaPrivateUddeim;
+
+/**
+ * Class PlgKunenaUddeIM
+ * @since Kunena
+ */
+class Easyblog extends CMSPlugin implements SubscriberInterface, DatabaseAwareInterface
```
xillibit: Should be Uddeim

Ruud68:
@xillibit fixed, and while doing so I also ran into the triggered events: onKunenaDisplay and onKunenaPrepare. I have created Events for these and refactored the triggers so they to are future proof.
Although onKunenaPrepare is triggered a lot, there is actually nothing in the repository (including add-on) that is listening this event?

Note also that the event has e.g. $this->topic as 'subject', but also $this->messages, this would not have worked as $this->topic is an object where $this->messages is an array: I ran into this while implementing the KunenaPrepareEvent which expects subject to be an object (just like the Joomla Core contentprepareevent (which it extends)

Both onKunenaDisplay and OnKunenaPrepare need to be tested if they still work. Hard for me as I do not have the means to test this myself (other then will it work on my environment without throwing errors

xillibit:
> Although onKunenaPrepare is triggered a lot, there is actually nothing in the repository (including add-on) that is listening this event?

I don't know neither where these events are used, it could be have been used at some point and removed.


~~I can't do nothing on the frontend have this error now, when trying to acces the forum :~~

xillibit:
I have removed everything and done a clean install, now by enabling the CB integration plugin : 
image

When enabling Uddeim integration plugin i can see that in backend under tools : 
image

Same error that above in backend but in frontend with Uddeim integration plugin enabled : 
image

Ruud68:
Ok, so this was the code:
```
$this->app->triggerEvent('onKunenaPrepare', ['kunena.topic.list', &$this->topic, &$params, 0]);
```
I refactored that into:
```
        $prepareEvent = new KunenaPrepareEvent('onKunenaPrepare', [
            'context' => 'kunena.topic.list',
            'subject' => $this->topic,
            'params'  => $params,
            'page'    => 0
        ]);
        $this->app->getDispatcher()->dispatch('onKunenaPrepare', $prepareEvent);
```

This is now more strict it it generates the error that you see because $this->topic doesn't exist in the function where it is used in the event.

So I commented out for now as this never worked.

xillibit:
Now with las changes with CB integration plugin : 
image

xillibit:
With Uddeim i have : 

Rendering Error in layout Widget/MenuBar: Class "Kunena\Forum\Plugin\Kunena\Uddeim\KunenaPrivateUddeim" not found in C:\Users\floflo\git\joomla-cms-5.4-dev\libraries\vendor\joomla\event\src\Dispatcher.php on line 454
Layout was rendered in C:\Users\floflo\git\joomla-cms-5.4-dev\components\com_kunena\template\aurelia\layouts\page\default.php on line 27

Showed here : 
image

Ruud68:
Uddeim should be fixed, was namespace issue.

The other error is not logical.
in ListDisplay line 157, KunenaParser::prepareContent($content, 'topicList_default'); is called, this in turn calls:
```
$content = HTMLHelper::_('content.prepare', $content, $params, 'kunena.contentprepare.' . $target);
```
So that line triggers onContentPrepare, and not onKunenaPrepare (which is what the error is about).

Can you check in you installation if this is also the case for you?

xillibit:
With your last changes i don't have this error with Uddeim, so with Uddeim i don't have any error now

Ruud68:
> Now with las changes with CB integration plugin : image

For the check I was referring to this one (comprofiler)

xillibit:
I had missed some files now with CB integration i don't have any error

Ruud68:
and the comprofiler 'onKunenaprepare' error is now also gone?

xillibit:
> and the comprofiler 'onKunenaprepare' error is now also gone?

yes it's gone

Ruud68:
Ok, so also added plg_kunena_finder as subscriber plugin

xillibit:
By comparing with Kunena 6.4.x when i enable CB integration, i don't have the avatar, the link to profle and the private link from CB. On K 6.4 : 
image

image


On K6.5 with our PR :  
image

Ruud68:
Hi @xillibit , can you provide me with a backup (akeeba) of your test site, so i can see what it is (not) doing? Without errors this is impossible to troubleshoot.

xillibit:
Hi, i have just send to you a mail with a link into it to download a backup

Ruud68:
Hi @xillibit, Found it (I hope). Was not an issue with the code, but with the setting.
an you disable avatar and profile in plugin_kunena_kunena?

Because you have set that as the avatar and profile provider as well, it is the first to respond and it will use that plugins avatar instead of the comprofile one.

Can you check that?

xillibit:
By disabling avatar and profile, i get avatar and profile link from CB. But it's not the same behaviour has K6.4, i don't have to disable avatar and profile in plugin_kunena_kunena to get avatar and profile link from CB : 
image

Ruud68:
On my production site I have a plugin for fetching the avatar / profile from ochBlog (basically user custom fields).
When I set plg_kunena_kunena avatar = yes, it will get the avatar from Kunena en not from my plugin, setting it back to no fetches the avatar from plg_kunena_ochblog again.

How triggered events work is that the code triggers the event onKunenaGetAvatar.
every plugin that listens to that event will answer, answering will be done in plugin order.

So the result is an array with all the provided avatars in the order of the plugin.

in 6.4 but also in 6.5 we use the first reported avatar in the array.

in 6.4, you can see that it will set the self::$instance to the first $class that is an object (avatar) >
```
            $classes = Factory::getApplication()->triggerEvent('onKunenaGetAvatar');

            foreach ($classes as $class) {
                if (!\is_object($class)) {
                    continue;
                }

                self::$instance = $class;
                break;
            }
```
this is for 6.5: $avatar is either the avatar or the array (when multiple plugins) of avatars in the order of execution, if it is an array (which it is because both kunena and comprofile give an avatar), it will set self::$instance to the first one $avatar[0]
```
            $avatarEvent = new KunenaGetAvatarEvent('onKunenaGetAvatar');
            Factory::getApplication()->getDispatcher()->dispatch('onKunenaGetAvatar', $avatarEvent);

            $avatar = $avatarEvent->getAvatar();

            if ($avatar instanceof KunenaAvatar) {
                self::$instance = $avatar;
            } elseif (\is_array($avatar) && !empty($avatar) && $avatar[0] instanceof KunenaAvatar) {
                self::$instance = $avatar[0];
            } else {
                self::$instance = new self();
            }
```

So I think that on your kunena 6.4 site the plg_kunena_kunena is ordered (or if not ordered, executed first) before the plg_kunena_comprofiler

xillibit:
I have tested on K6.4 by changing the order of the plg_kunena_kunena and so you are right when i put in first it doesn't get the CB avatar and when it's in last it get the avatar from CB

Ruud68:
And while investigating that I found two other issues:
we have events that should only return one result: avatar, private, profile
we have events that should return multiple results: accesscontrol, activity, login

So or the single events, I have added a stoppropagation, this to avoid the second triggered plugin overwriting the first triggered plugin value: this then resulted in e.g. the avatar of the last plugin being displayed first.

I have refactored the 'multiple' events so the triggered event now gets an array with all the objects in it: it can then use all or decide which one to use, I crawled through the old code and this is the same now.

So all should be okay now, except for the easysocial plugin as that is not yet converted to a subscriber / service provider plugin.

PR #9917, opened 2025-09-22
eliminate need for user (stopforumspam) plugin

Ruud68:
Pull Request for Issue # . 
in the add-on repository we have a user plugin that has one function: check a new user against the stopforumspam API on username / email / IP-address
We can eliminate this plugin altogether if we add that logic to the plg_system_kunena plugin.
 
#### Summary of Changes 
stopforumspam new user check added to system plugin. It can be enabled via the Kunena Configuration > tab Security > heading 'Stop Forum Spam Configuration' > Check New User (yes / no). Default is no.
 
#### Testing Instructions
1. remove (if installed) the plg_user_kunena plugin
2. enable new User Check in Kunena Config
3. create new user account (front-end) with username 'spammer': this will be reported back by StopForumSpam api as spammer.

An error will be thrown: "Registration failed. Please try again or contact the site administrator for assistance."


xillibit:
I tried but i get too this other message saying that an email has been sent with a verification link : 
image

Ruud68:
Hi @xillibit,
when registering with a spam account (e.g. username 'spammer'). The registration is aborted as an error is thrown ('Registration failed. Please try again or contact the site administrator for assistance.').

The messages you get are when the registration is 'successful', so not stopped by the stopforumspam part.

so the test should be two fold:
1. register user with username 'spammer' > error thrown and user NOT created
2. register other user name > user created and message / email / what is needed for registration on Joomla executed

The logic for the plugin (handling what to do in case of spammer / no spammer) has not changed logically, so this should work as it worked before using the separate user plugin

xillibit:
Re-tested it works like intented

PR #9918, opened 2025-09-23
Changing a category that has a Topic Template causes an error in the editor

zinshuke:
**Issue:** Changing a category that has a `Topic Template` causes an error in the editor.

**Steps to Reproduce:**

1. Use core Kunena.
2. Allow users to change the category when creating a new topic.
3. Open the **New Topic** page.
4. Make sure some categories have a `Topic Template`.
5. Switch between those categories.
6. Confirm replacing the content.
7. Notice the incorrect string displayed in the editor (see screenshot below).

image  

**Summary of Changes:**

* The AJAX request for `Topic Template` returns with a `data` property.
* This property needs to be added to the handler.


rich20:
- [x] Success

PR #9924, opened 2025-09-26
Refactor Config in Core Joomla Configuration

Ruud68:
Pull Request for Issue # . 
 
#### Summary of Changes 
This is complete refactor of the configuration for com_kunena. It has been completely removed and configuration settings have been converted to Core Joomla Component options.
The usage of the configuration settings in code is (almost fully) unchanged: Library file KunenaConfig.php which is responsible in the code for getting the configuration has been completely rewritten, but the 'interface' with it is still the same
 
#### Testing Instructions
1.  BACKUP!!!
2. when reporting issues, please report if it works correct in the old version and is introduced in this version. If it also didn't work in the old version then please create a separate bug report.
3. confirm that upgrade from version lower then 6.4.5 will not be possible 9installation of the package should give an error and exit.)
4. Import + conversion of Configuration: after installation of 7.0.0.DEV are the configuration setting the same as they where. Special attention on fields where you select categories as these have been converted in format (from string to array))
5. check if ./administrator/components/com_kunena [tmpl/config | src/View/Config | src/Controller/ConfigController.php | src/Model/ConfigModel.php] have been removed
6. Back-end: is configuration available via menu 'Configuration' and 'options' buttons in the topbars
7. For speed, configuration settings are cached (in cache, cache group = com_kunena_configuration). When saving the configuration this cache is removed by the system plugin so new settings are direclty available. does this work correct?
8. front-end: visit all menu's, links, tabs: if you see an error 500: 'Trying to set non-existing property: [xyz]', or 'Trying to get non-existing property: [xyz], please report here with trace and error message
9. check if when configured to display certain categories in a view, this works correct
10. check if RSS still works correct, test with different category config settings (include / exclude)
11. developer: check in your IDE is autocompletion for the KunenaConfig properties works correct. note there are only two types: string and array as that is the only thing the the parameters return.

#### Re-importing settings:
the upgrade to 7.0.0.DEV doesn't remove the old configuration table, on installing (upgrading) 7.0.0.DEV, the table contents is converted and imported into the #___extensions table record for com_kunena, field params. This is done one time. If you want to test it again, then you need to set the params for com_kunena in the #__extensions table to '{}'. Installing the 7.0.0.DEV package will then import the configuration again.

Upgrade to 7.0.0.DEV is only possible when the installed version is 6.4.5 or higher: this is because the conversion of the configuration depends on the correct configuration variable names, these have been changed during the releases so we want to make sure we have the correct set

#### Removed:
The complete Config MVC + tmpl has been removed

#### Changed properties:
latestCategory, rssExcludedCategories, rssIncludedcategories previously returned a string (comma separated categorie ids), now they return an array. I have changed the TopicController to handle this change.

@xillibit : the following settings are static (they are set by code in 6.4.5 and not via configuration), is this correct or am I missing something?

- disableEmoticons = 0
- enableThreadedLayouts = 0
- allowEditPoll = 1
- maxPersonalText = ''


rich20:
I have tried it on several test installations, but I am unable to install Kunena with this pull request.


<blockquote class="blockquote">
                            <span class="badge bg-secondary">0</span>
                            Call to undefined method Joomla\Filesystem\Folder::exists()                        </blockquote>
                                                    <div>
                                <h2 id="caption">Call Stack</h2>

 **| Function | Location**
-- | -- | --
1 | () | JROOT/tmp/install_68d70df8596ed/K7.0-2025-09-27/script.php:485
2 | Joomla\CMS\Installer\InstallerScriptInterface@anonymous/.../tmp/install_68d70df8596ed/K7.0-2025-09-27/script.php:38$e0->deleteFilesFolders() | JROOT/tmp/install_68d70df8596ed/K7.0-2025-09-27/script.php:478
3 | Joomla\CMS\Installer\InstallerScriptInterface@anonymous/.../tmp/install_68d70df8596ed/K7.0-2025-09-27/script.php:38$e0->deleteRemovedFilesFolders() | JROOT/tmp/install_68d70df8596ed/K7.0-2025-09-27/script.php:189
4 | Joomla\CMS\Installer\InstallerScriptInterface@anonymous/.../tmp/install_68d70df8596ed/K7.0-2025-09-27/script.php:38$e0->preflight() | JROOT/libraries/src/Installer/Adapter/PackageAdapter.php:593
5 | Joomla\CMS\Installer\Adapter\PackageAdapter->triggerManifestScript() | JROOT/libraries/src/Installer/InstallerAdapter.php:714
6 | Joomla\CMS\Installer\InstallerAdapter->install() | JROOT/libraries/src/Installer/Installer.php:652
7 | Joomla\CMS\Installer\Installer->install() | JROOT/administrator/components/com_installer/src/Model/InstallModel.php:215
8 | Joomla\Component\Installer\Administrator\Model\InstallModel->install() | JROOT/administrator/components/com_installer/src/Controller/InstallController.php:52
9 | Joomla\Component\Installer\Administrator\Controller\InstallController->install() | JROOT/libraries/src/MVC/Controller/BaseController.php:730
10 | Joomla\CMS\MVC\Controller\BaseController->execute() | JROOT/libraries/src/Dispatcher/ComponentDispatcher.php:143
11 | Joomla\CMS\Dispatcher\ComponentDispatcher->dispatch() | JROOT/libraries/src/Component/ComponentHelper.php:361
12 | Joomla\CMS\Component\ComponentHelper::renderComponent() | JROOT/libraries/src/Application/AdministratorApplication.php:150
13 | Joomla\CMS\Application\AdministratorApplication->dispatch() | JROOT/libraries/src/Application/AdministratorApplication.php:205
14 | Joomla\CMS\Application\AdministratorApplication->doExecute() | JROOT/libraries/src/Application/CMSApplication.php:304
15 | Joomla\CMS\Application\CMSApplication->execute() | JROOT/administrator/includes/app.php:58
16 | require_once() | JROOT/administrator/index.php:32

</div>

Ruud68:
@rich20 , so this error comes from the package install script, this issue now surfaces because this function is now used again and on previous updates not.
I have undone my changes, and created an install script for the component.

Can you rebuild and check if this now works correct? (note, also the build file has changed)

@xillibit because in the package script we now only allow updates from 6.4.5, this gives the opportunity to remove all old updates / checks / installs / conversions from the package script as these will never be executed. Just like Joomla when going to 6 it requires being on 5.4 this enables starting with a clean slate.

rich20:
Now i get:
- [x] Kunena Configuration settings were automatically converted to the new Configuration settings, please validate after installation completes.
- [x] The configuration settings remain
- [x] Confrmed, upgrades from older versions such as K6.4.5 are not possible. Message: " You cannot not upgrade Kunena from version 5.2.9, you can only upgrade from version 6.4.5 or higher"

To what extent does it still comply with the test specifications with the new script? For example, point five “check if ./administrator/components/com_kunena [tmpl/config | src/View/Config | src/Controller/ConfigController.php | src/Model/ConfigModel.php] have been removed.” No files are removed.

Ruud68:
Thanks @rich20 , I checked on upgrade where that should be on update when removing the files / folders. Should be fixed now.

rich20:
- [x] Folder | File removed
- [x] Frontend tabs work
Further tests to follow

rich20:
Summary:
- [x] Kunena Configuration settings were automatically converted to the new Configuration settings, please validate after installation completes.
- [x] The configuration settings remain
- [x] Confrmed, upgrades from older versions such as K6.4.5 are not possible. Message: " You cannot not upgrade Kunena from version 5.2.9, you can only upgrade from version 6.4.5 or higher"
- [x] Folder | Files removed - ( [tmpl/config | src/View/Config | src/Controller/ConfigController.php | src/Model/ConfigModel.php])
- [x] Frontend tabs work
- [x] Feed include | exclude works when you use the general feed button at the bottom.
However, if a category has been excluded, you can still access it via the index (see image). But this is a bug that has been around for a long time.
.
image910

xillibit:
> @xillibit : the following settings are static (they are set by code in 6.4.5 and not via configuration), is this correct or am I missing something?
> 
>     * disableEmoticons = 0
> 
>     * enableThreadedLayouts = 0
> 
>     * allowEditPoll = 1
> 
>     * maxPersonalText = ''

Except for maxPersonalText which are in user setting the others ones are present in KunenaConfig in K6..4.5

> @xillibit because in the package script we now only allow updates from 6.4.5, this gives the opportunity to remove all old updates / checks / installs / conversions from the package script as these will never be executed. Just like Joomla when going to 6 it requires being on 5.4 this enables starting with a clean slate.

I didn't see this change. Because now it have set to K7.0 some old uptate parts can be removed

Ruud68:
@xillibit, currently the disableEmoticons property is set to 0 (in 6.4.5), you cannot configure it, or at least when searching the repo for this variable i can only find istances where it is used, not where it can be toggled on / off by the administrator.

same for enableThreadedLayouts, allowEditPoll, and maxPersonalText
in the config model there are yesno lists created for the selection but these where never used in the tmpl/config/default.php where the actual configuration is done.

I have now added them to config.xml as static (read: hidden) to avoid having errors when trying to read these properties.
Should they be configuarble by the user?

You say maxPersonalText is a User tab setting, but i only see maxSig there (which I think is not the same).

xillibit:
> I have now added them to config.xml as static (read: hidden) to avoid having errors when trying to read these properties.
> Should they be configuarble by the user?

Let them non-configurable by user like it was, i don't remember the reason why it wasn't configurable by user

xillibit:
With this PR, I can't restore Kunena menu : 
image

Ruud68:
@xillibit, just added property defaultPage = recent. another static one that was not available via configuration in 6.4.5
let's find them all :)

rich20:
> currently the disableEmoticons property is set to 0 (in 6.4.5), you cannot configure it, or at least when searching the repo for this variable i can only find istances where it is used, not where it can be toggled on / off by the administrator.

This option disable/enableEmoticons used to be found on the BBCode tab, but was removed a long time ago.

> same for enableThreadedLayouts, allowEditPoll, and maxPersonalText
in the config model there are yesno lists created for the selection but these where never used in the tmpl/config/default.php where the actual configuration is done.

This option ThreadedLayouts and other layouts was available in the early days of Kunena, but has never been used in newer versions. We have not removed this setting only so that other template authors can use it. These options have no function in our template.

> allowEditPoll

This setting is located under the “Extra” tab (at the very bottom), but a yes/no button would also suffice.

> You say maxPersonalText is a User tab setting, but i only see maxSig there (which I think is not the same).

You're right, it's not the same. There is no length setting for the personal text (at least I can't find such an option). But even in Kunena 6.4, the maximum length is 50 characters; I can't enter more than that.

Ruud68:
Good morning @rich20 , valuable info!

so with that information I would propose the following:

1. replace property allowEditPoll with allowUserEditPoll (as that is what the setting is in the configuration > this is not used)
2. disableEmoticons > move this to template parameters as that is where it is used, default to 0 when not configured
3. maxPersonalText > move this to template parameters as that is where it is used, default to 50 when not configured
4. defaultPage > leave as is (fixed to 'recent')
5. enableThreadedLayouts > move this to template parameters as that is where it is used, default to 0 when not configured

1 = done
2,3,5 I have refactored to use the template parameters, as these are not set in aurelia the default values will be used. Template developers can then add them to their template params and set the value accordingly.

rich20:
'Fix for removing Kunena via Tools' Has the option been completely removed?

xillibit:
Yes this option has been removed

Ruud68:
This PR now also has the fix for the complete removal of the media/kunena folder when uninstalling the package

Ruud68:
> Good morning @rich20 , valuable info!
> 
> so with that information I would propose the following:
> 
>     1. replace property allowEditPoll with allowUserEditPoll (as that is what the setting is in the configuration > this is not used)
> 
>     2. disableEmoticons > move this to template parameters as that is where it is used, default to 0 when not configured
> 
>     3. maxPersonalText > move this to template parameters as that is where it is used, default to 50 when not configured
> 
>     4. defaultPage > leave as is (fixed to 'recent')
> 
>     5. enableThreadedLayouts > move this to template parameters as that is where it is used, default to 0 when not configured
> 
> 
> 1 = done 2,3,5 I have refactored to use the template parameters, as these are not set in aurelia the default values will be used. Template developers can then add them to their template params and set the value accordingly.

BTW, this is a B/C break for templates: these should update their code. So probably good to note this when we do a RC version?

xillibit:
@Ruud68 : i just noticied that the scripts are now under /media/kunena/kunena/ instead of /media/kunena/, by using the powrshell script to make symlinks then i have plenty of error. Like : 

`PHP Warning:  DOMDocument::load(): I/O warning : failed to load external entity &quot;file:/C:/Users/floflo/git/joomla-cms-6.1-dev/media/kunena/core/svg/file.svg&quot; in C:\Users\floflo\git\Kunena-forum-K7.0-fork\src\libraries\kunena\src\Icons\KunenaSvgIcons.php on line 66`

This change has been made in this PR

Ruud68:
Hi @xillibit , I back tracked it to this one: https://github.com/Kunena/Kunena-Forum/pull/9924/changes/1279436617db4c0d356a40d8632a2743ac0b3d1b

I have no idea (anymore) as to why that changed, doesn't seem logical and IMO should not have happened.

When reverting the file renames, it is best to test install of the package, and also uninstall as this is what this change was about.
If both work then we can be sure that the file moving was unintentional.

rich20:
@Ruud68
I can't remember why you changed that either, but I know you explained at the time why you did it.
In any case, the path /kunena/kunena must still exist somewhere, because I'm getting this message now and the /media/kunena folder remains empty after installation.
Warning
JInstaller: :Install: File does not exist [ROOT][TMP]/install_69f869b599517/K7.0.5-DEV/packages/install_69f869b622d8b/kunena_media/kunena/kunena

Ruud68:
@rich20 I created a PR #10060 to address this and also to make sure the uninstall is working as that was in the first place the issue we solved by moving the directory.

Issue #5427, opened 2017-06-11
Feature Request: Scheduled Trashbin Purge


GJSchaller:
Currently, the Trash Bin needs to be purged / emptied manually.

I would like to suggest a task that can be scheduled, either by plugin or by CRON job, that would remove all deleted posts older than (X) days, as set by an Admin.

This would mean that once a day / week / month, deleted topics more than 7 / 14 / 30 days old would be automatically purged, making administration much easier and keeping the forums clean.

xillibit:
A plugin which handle that by a cron job should be the better solution

Ruud68:
We can now implement this using the new (K7.0) task scheduler plugin. What do you think: a viable option?
When time permits I can set this up

GJSchaller:
This is an older request, but the more automation we can do, the less human effort is needed to maintain things.  I would very much like this.

Ruud68:
@GJSchaller : progress :) if you are able to test that would be great!

GJSchaller:
Unfortunately, the community I used to support hasn't used Kunena in a few years, I don't have the environment to test this any more.

Ruud68:
This one can be closed: part of Kunena 7.0 :)

PR #9954, opened 2025-11-10
New: scheduled task purge trashbin

Ruud68:
Pull Request for Issue #5427  . 
 
#### Summary of Changes

- moved login for deleting topics / messages out of the controller into the model (so we can use it in multiple places
- added deleted_time to #__kunena_messages table, this will hold the time of deletion so we can use that
- added new task in kunena task scheduler plugin > you need to add it manually and can configure the number of days to purge the trashbin on (so 7 days, will purge all messages that have been deleted more then 7 days ago)

 
#### Testing Instructions

- Install new 7.0.0 build
- make sure to go to the kunena dashboard, so table get updated
- delete a message in the front-end, delete a topic in the front-end
- in backend > kunena > Trashbin:
  - test if purging of messages is working (one, multiple)
  - test if purging of topic is still working (one, multiple)
- create scheduler task for Kunena purge
- delete messages / topics in the front-end:
  - check that BEFORE the number of configured days the messages a/ topics have NOT been purged from the trashbin
  - check that AFTER the number of configured days the messages a/ topics have been purged from the trashbin


rich20:
The column deleted_time is not created. Could this be because the command is CREATE TABLE IF NOT EXISTS `#__kunena_messages`? However, the table #__kunena_messages already exists in an existing installation; only the column deleted_time needs to be inserted.

I have now created the column with phpMyAdmin, but I can't find anything in the backend that allows me to set the time for removal (permanent deletion).

Ruud68:
> The column deleted_time is not created. Could this be because the command is CREATE TABLE IF NOT EXISTS `#__kunena_messages`? However, the table #__kunena_messages already exists in an existing installation; only the column deleted_time needs to be inserted.

Hi, it should have added that column upon visiting the dashboard after installation, at least it did for me. But it checks for the version, so maybe you already had a version of 7.0.0.BETA3 installed and that is the reason why it didn't add it.


> I have now created the column with phpMyAdmin, but I can't find anything in the backend that allows me to set the time for removal (permanent deletion).

You do not set the time for removal, the delete_time is set with the time of message deletion: so when you click delete, then the deleted time will be the time of the click.

does this make sense?

rich20:
Sorry, it was my mistake, I already had K7.0.0-BETA3 installed. I have now tested an upgrade from K7.0.0-DEV, and the database upgrade works.
Manual deletion and cleanup work as intended, but unfortunately I can't test the rest because I've never created a scheduler task or CRON job and don't know how to do it.

Ruud68:
Good to hear it also works for you (adding the field).
as for the cron task:

1. in back-end > System > Scheduled Tasks: you will see a list of already scheduled tasks (like: Kunena - Remove Expired Bans)
2. click [New] > Select a task type you want to add, in this case: Kunena Purge Trashbin
3. Fill in:
  - Task name
  - Execution Rule (you can set that e.g. to manual (you can then run the task manually)
  - Age in Day: the number of days for the deleted messages / topics. So 31 means 31 days after a message was deleted it gets purged
  
That's it!
Now by default the task scheduler runs as a so called 'Lazy Scheduler', that means that the tasks are checked if they are up for execution upon a visit on your front-end.
That is why on a development machine where I have no visitors I usually test with execution rule Manual.

rich20:
Thanks for the explanation! I tested it. It now works both manually and when I set the schedule in minutes. However, if the setting is “Age in days = 0,” the entire trash is emptied.
I can only determine after three days whether only older posts are being removed as desired by setting the time to two days and delete today and several posts tomorrow.

rich20:
- [x] Everything works as it should, including the scheduler.

Issue #9949, opened 2025-10-28
Error 500 shown on all  invalid/outdated URLs in Kunena Forum


c-schmitz:
**Describe the bug**
When opening a URL where the post does not exist (for whatever reason) a 500 error is thrown, instead of a 404.
This issue is problematic for SEO reason as 500 errors cause penalties with search engine.

**To Reproduce**
1. Go to (https://www.kunena.org/forum/general-user-contributiions/168985-kunena-forum-menu-as-template)
2. See error 500

**Expected behavior**
I would expect a normal 404

**Actual result**
A 500 error is shown

Joomla version: 5.4.0
Kunena version: .5
Php version: PHP 8.1
Database version: MariaDB

c-schmitz:
That was a hard nut to crack - but found it.

Ruud68:
@greenhavendevelopersmarketing-stack please test #9969

c-schmitz:
@Ruud68 It is just spam.

Ruud68:
Hi @xillibit thanks for merging, as this is a 'router' change, this really (!) needs to be tested thoroughly. This because changes to the router have high impact as it is complex and used in my different use cases.
I don't feel this change had enough tests, maybe I'm to careful, but as said: the Joomla router is called router-from-hell for a reason :)

xillibit:
I will install this evening a nightly build of K7.0.2-DEV with our changes on k.org and i will test it too after

PR #9950, opened 2025-10-28
Fixed issue #9949: Error 500 shown on all invalid/outdated URLs in Kunena Forum 

c-schmitz:
Fixed issue: Deprecation warning "Creation of dynamic property Kunena\Forum\Libraries\Forum\Category\KunenaCategory::$image is deprecated"

Pull Request for Issue #9949 
 
#### Summary of Changes 
Fixed the 500 errors in on invalid URLs and shows a proper 404 now
Fixed deprecation error shown on category listing

PR #9969, opened 2026-01-02
attempt to fix error 500 when URL not correct

Ruud68:
Pull Request for Issue #9949 . 
 
#### Summary of Changes 
When you have a correct URL like: [domain]/forum/category/123-topic-alias where forum is the com_kunena home page, category is a valid category alias and 123 is a valid topic id and you do:
1. change the topic id to a non-existing topic id number
2. change the category alias to a non existing category alias

you will get a:
1. 404 error, but with deprecated messages
2. 500 error

This PR fixes:
1. the deprecated messages
2. when changing to a non-existing category it will now give a correct 404 instead of a 500

What this PR fixes for 2. is that it 'breaks' out of the logic when the category doesn't exist instead of setting the view to category and then use the topic id (123) as category id.

#### Testing Instructions
now this is where it gets tricky as I think we need to get as meny test cases as possible: the category can be wrong, but can also have a ghost category, or subcategories, etc. My test site is verly basic when it comes to sections / categories and I never set a ghost category when doing e.g. a rename of a category etc (not even sure  that is possible).

@c-schmitz can you check if your test cases (could not find these in your original topic) are fixed with this change (after you revert your own fix for testing)

### DIFF
--- a/src/site/src/Service/Router.php
+++ b/src/site/src/Service/Router.php
@@ -217,11 +217,12 @@ class Router extends RouterView
             $id = (int) $query['id'];
 
             if ($id) {
-                $subject = KunenaRoute::stringURLSafe(KunenaTopicHelper::get($id)->subject);
+                $subject = KunenaTopicHelper::get($id)->subject;
 
                 if (empty($subject)) {
                     $segments[] = $id;
                 } else {
+                    $subject    = KunenaRoute::stringURLSafe($subject);
                     $segments[] = "{$id}-{$subject}";
                 }
 
@@ -382,9 +383,9 @@ class Router extends RouterView
             $var   = array_shift($seg);
             $value = array_shift($seg);
 
-            if (empty($var) && empty($value)) {
+            if (empty($var) || empty($value)) {
                 // Skip /-/
-                continue;
+                break;
             }
 
             if (is_numeric($var)) {
--- a/src/site/src/Service/Router.php
+++ b/src/site/src/Service/Router.php
@@ -373,6 +373,8 @@ class Router extends RouterView
                     $sefcats = false;
                     $vars    = $variables + $vars;
                     continue;
+                } else {
+                    break;
                 }
             }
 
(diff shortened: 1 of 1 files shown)


c-schmitz:
Seems to work fine.

Ruud68:
@c-schmitz so your previous comment that it didn't work was not valid and it does work? Just to make sure I'm looking at the same issue here :)

c-schmitz:
Yes, that's why I deleted the previous comment.

c-schmitz:
My fix from #9950 is still valid though, because the way the exception is re-thrown right now at that code location is not working for any exception - it would always throw just a 500. 
From my understanding you are just catching the error earlier in the flow now. That's why the Joomla generic 404 shows, not the Kunena one.

Ruud68:
The router (where I did the fix) is responsible for parsing a sef URL into the parts needed to goto the component, view, layout with the appropriate data > id, catid, etc.

What happened is that when the sef URL was wrong (e.g. non existing category or topic), the router had a bug and and set the topic id in the url as category id in the data: that would give an error when trying to read this non-existing category.

So this was actually a bug that I fixed.
So now when the category in the url is non-existing the router will do its default (joomla) behavior: give a 404 as that category doesn't exist.
Same for the topic: if you change the topic id in the url, it should do the default joomla router behavior: give a 404 as that url is pointing to a non-existing page.

your change fixed the 'result' but didn't fix the (root) cause.

c-schmitz:
Yes, that's what I thought, which is fine and I agree. 
Just wanted to suggest maybe still to apply my fix too, for other situation where that exception might be called.

c-schmitz:
After having your patch applied for a while I noticed errors happening. 
Trying to edit a user from the Kunena frontend or moderating a post results in a 404.

Ruud68:
@c-schmitz good catch, can you try again with the additional changes I just pushed?

c-schmitz:
Seems to work better this time.

makitso:
The change to Router.php results in the forum attachments to fail.
When you open a post with an attachment in the body of the post, clicking on it gets error.  Reverting to previous code in Router.php fixes the problem.

Issue #10017

Ruud68:
@makitso, just to confirm that this is caused by deprecated messages thrown, can you set error reporting to none in Joomla and turn debug to off?
If that is already the case then let me know so i can investigate further

makitso:
Error reporting off as well as debuging -- nochange to the problem.

Issue #9974, opened 2026-01-16
Config parameters lost during update from K6.4 to K7.0


xillibit:
I have done an updagred from K6.4.8 to K7.0 (build of today) on kunena.org, after going to Kunena dashboard it has done the following updates steps : 

- 7.0.0-DEV
- 7.0.0-BETA3

I have discovered that some settings has back to their values, not checked all yet but some examples : 

- Tab Frontend -> Show Avatar on Category Index and Recent Topics 
- Tab Upload -> Maximum Image File Size (KB)
- Tab Upload ->  Maximum File Size (KB)

@Ruud68 : when you had done refactoring of config part, is-there some changes in their which can casue that issue ?

Ruud68:
Hi @xillibit ,
conversion is done here: https://github.com/Kunena/Kunena-Forum/blob/8072e6bbb3a9218311163e003e94911efac36a17/src/admin/script.php#L94

it is run on postflight, it checks if Kunena is already installed (update) and if the installed version < 7.0.0
Then it read the params from #__kunena_configuration (where id = 1)

These fetched params are 'preprocessed' for latestcategory, rssExcludedCategories, rssIncludedCategories. This is done because these specific params are save in in old config as string and not as arrays.

The processed params are the stored in the #__extensions table for the kunena component.

ALL params are stored and should be retained, but if you open the configuration it will read the newly stored params into the configuration form. it will use the field name in the config.xml and read that value from the params.

There can only be an issue when there is a field in the config XML that is NOT in the 6.4.8 params (or is there with another name). When you now save the configuration, only the fields in the config.xml will be saved into the #__extensions table.

So with this logic: are the missing values after upgrade from 6.4.8 to 7.0.0 due to a different or missing field name in config.xml?

that is something you could check.

xillibit:
I have checked in table #__kunena_configuration i have the params : 

- "avatarOnCategory":"1"
- "fileSize":"6096"
- "imageSize":"6096"

In the config.xml there are the same : 

- avatarOnCategory
- fileSize
- imageSize

I have checked what it done on admin/script.php, the params quoted above are right here with the right value

I have found there are others values which have lost : email, emailHeader, latestCategoryIn, latestCategory, stopForumSpamKey, ebayApiKey, googleMapApiKey

xillibit:
After testing a new time an upgrade from K6.4.8 to K7.0.0 my fix does the job

Ruud68:
[...]

PR #9980, opened 2026-01-26
fix script + not use old config values

Ruud68:
Pull Request for Issue #9974 . 
 
#### Summary of Changes 
 revert change #9974 
add cleaning of old / unused configuration values before updating

#### Testing Instructions
1. install on K6.4.8, check if all config settings for Kunena have been imported correct,
2. check if other extensions still have correct configuration 

@xillibit, not sure if the added logic will fix your issue where some config settings where not imported correct, if not I advice not to merge this PR and just revert your own PR.

Note that your PR didn't change anything on the imported configuration, so if the data was correct with your PR it should have also been correct without your PR: the only thing your PR changed where the data was saved to

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -632,30 +632,29 @@
         },
         {
             "name": "doctrine/instantiator",
-            "version": "2.0.0",
+            "version": "2.1.0",
             "source": {
                 "type": "git",
                 "url": "https://github.com/doctrine/instantiator.git",
-                "reference": "c6222283fa3f4ac679f8b9ced9a4e23f163e80d0"
+                "reference": "23da848e1a2308728fe5fdddabf4be17ff9720c7"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/doctrine/instantiator/zipball/c6222283fa3f4ac679f8b9ced9a4e23f163e80d0",
-                "reference": "c6222283fa3f4ac679f8b9ced9a4e23f163e80d0",
+                "url": "https://api.github.com/repos/doctrine/instantiator/zipball/23da848e1a2308728fe5fdddabf4be17ff9720c7",
+                "reference": "23da848e1a2308728fe5fdddabf4be17ff9720c7",
                 "shasum": ""
             },
             "require": {
-                "php": "^8.1"
+                "php": "^8.4"
             },
             "require-dev": {
-                "doctrine/coding-standard": "^11",
+                "doctrine/coding-standard": "^14",
                 "ext-pdo": "*",
                 "ext-phar": "*",
                 "phpbench/phpbench": "^1.2",
-                "phpstan/phpstan": "^1.9.4",
-                "phpstan/phpstan-phpunit": "^1.3",
-                "phpunit/phpunit": "^9.5.27",
-                "vimeo/psalm": "^5.4"
+                "phpstan/phpstan": "^2.1",
+                "phpstan/phpstan-phpunit": "^2.0",
+                "phpunit/phpunit": "^10.5.58"
             },
             "type": "library",
             "autoload": {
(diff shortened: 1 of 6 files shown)


rich20:
> Testing Instructions
>    install on K6.4.8, check if all config settings for Kunena have been imported correct,
>  check if other extensions still have correct configuration

I tested it with Joomla 5 and Joomla 6. Both times, the configuration remained intact.
Since I don't run a website myself, extensions are in short supply for me. But with CommunityBuilder, the settings remained intact, as did the settings for Kunena extensions such as Kunena Discuss, etc.

Messages after installation:
> Kunena Configuration settings were automatically converted to the new Configuration settings, please validate after installation completes.
>Successful updated.

>Datenbank aktualisiert zu Version 7.0.0-DEV
>info
>PLG_KUNENA_COMPROFILER_WARN_VERSION

xillibit:
I have tested this change on a backup of k.org with K6.4.8 at the end of installation i have only that : 
image

In configuration of Kunena i have the same value lost like here : https://github.com/Kunena/Kunena-Forum/issues/9974#issuecomment-3796242754

Ruud68:
@xillibit I see in the screenshot that the import didn´t run, so nothing imported from the old settings.
The check is twofold:
1. only execute when upgrading from 6.4.8
2. only execute when #__extensions entry for com_kunena has empty params field (really empty, so not {}).

My guess is that 2. is not empty and therefore it doesn't import the settings

rich20:
> The check is two fold:
> 1. only execute when upgrading from 6.4.8

If the upgrade can only be performed correctly from version 6.4.8 onwards, we should modify the script.php file in Kunena 6.4.8 so that K6.4.8 can also be installed on Joomla 6.x.
Most users first update Joomla to 6.0, und it remain an old Kunena version on their installation that can no longer be updated to K6.4.8. They then try to update directly from versions such as K6.2, K6.3 to K7.0.

Ruud68:
@rich20 good point!

xillibit:
@Ruud68 : i have just set up the backup of k.org with K6.4.8, in the table #__extensions the column params is already filled by something for the line com_kunena. The content of params is talking about K2.0.4 : 
image

Is-there a way to check the content if it's not JSON for Kunena configuration and make a copy of the line just in case before to get ride the content of params ?

@rich20 : at the begining at the install of K6.4.8 it should check if the Joomla! compatiblity plugin is installed to be sure. I will need to apply changed made on script.php from K7.0 to K6.4 and then make a new release K6.4.9

I had added a check which prevent to install K6.4.8 from 6.0.0-alpha1 and beyond, i was thinking to change that to 6.1.0-alpha1

Ruud68:
@xillibit , so that is it then: params not empty...
Tomorrow I will do a check on params if not empty or if a variable in params is present.
Good catch! I have a site dating back to kunena 3 where params was just empty, so is a good test case.

Ruud68:
@xillibit okay so I have updated this pr:

1. it now correctly adds the component id to the where update query avoiding the overwriting of all component params
2. removed the check if params is empty or not: it will now always import the data when installed version of Kunena is < 7.0.0. The reason why The check was in there was for not overwriting the params when going e.g. from 7.0-beta to 7.0-rc as both are < 7.0.0. Now that we are done with these releases we do not need that check anymore. So in your case params is not empty but will be overwritten anyway because you are comming from 6.4.8 ( < 7.0.0)

xillibit:
There is a conflict on src/admin/script.php

I will give a try in the evening on a backup with K6.4.8

Ruud68:
@xillibit I just did a rebase, so conflict should be solved

xillibit:
Works as expected on a backup of k.org with K6.4.8 : 
image

PR #9981, opened 2026-01-26
Change: config save now always saves when called even when no changes…

Ruud68:
… detected

Pull Request for Issue #https://www.kunena.org/forum/kunena-7-0-0-support/169043-unable-to-change-default-template . 
 
#### Summary of Changes 
change in configuration not detected correct in certain installations
 
#### Testing Instructions
switch default template and see if the switching works and is 'retained', previously it could be that it was not possible to switch.

### DIFF
--- a/src/libraries/kunena/src/Config/KunenaConfig.php
+++ b/src/libraries/kunena/src/Config/KunenaConfig.php
@@ -584,7 +584,6 @@ class KunenaConfig
     public function save(): bool
     {
         $params  = ComponentHelper::getParams(self::COMPONENT);
-        $changed = \false;
 
         foreach ($params as $key => $value) {
             if ($this->config[$key] === $value) {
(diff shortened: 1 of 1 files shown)


rich20:
I tested this fix, it doesn't produce an error, but unfortunately I don't have a second template to test the switching.

cssyeah:
[ktsample.zip](https://github.com/user-attachments/files/24872720/ktsample.zip)
Hi @rich20 , You can use the example template I've attached to test this error. I've tried many times but still can't set the default template. I cloned this template from the default Kunena component template to ensure everything works correctly.

Thanks

Ruud68:
@cssyeah can you try the following:
1. in the template overview: click the 'other' template link (bullit) to make it default
2. confirm that it is not displaying the template as 'default'
3. click the [options] button (top-right)
4. do not change any thing but click [save & close] 
5. check if the default template is now correctly displayed to the one you selected in step 1.

cssyeah:
@Ruud68 Thank you very much.

I followed some of your steps. It worked. However, is this an issue that needs fixing in the next version of the Kunena component?

Thanks again.

Ruud68:
@cssyeah thanks for confirming. I now also found a server where I could reproduce this issue and did a fix for it in my last commit here.
Can you test this one as well as this will be the permanent fix (fingers crossed) if it works.

logic @xillibit :
so we cache the kunena configuration into a joomla cache file (com_kunena_configuration.php), when you change the template we push this change into the component parameters and because that changed we also need to remove the com_kunena_configuration.php file (which we did), but.... it turns out that Joomla now also caches the components configurations: so when we recreated the cache and asked Joomla for the stored params, we actually didn't get the stored params (with the new default template), but the joomla cached params (with the old default template).

This change now not only clears the com_kunena_configuration cache file but also the _system cache file (where joomla stores the parameters).

rich20:
@cssyeah Thank you very much for the test template.
@Ruud68 I can also confirm that everything works. I also tested various configurations between the templates to see whether settings from one template affect the other template. This is also not the case; everything is fine.

Ruud68:
@rich20 thanks for testing, appreciate it! Just to make sure: are you testing with the last commit I did in this PR?

rich20:
I mean, these commands loaded your pull request correctly.
git fetch origin pull/9981/head:conf
and
git checkout conf
And with that, I created a build.

PR #9993, opened 2026-02-12
Change: execute database upgrades from install script instead of via …

Ruud68:
…dashboard

Pull Request for Issue # all issues related to people not going to Kunena Dashboard after install of new package. 
 
#### Summary of Changes 
This PR executes database updates from withing the install script: this to ensure that the updates are always executed instead of being executed via the Kunena Dashboard (which a lot of people forget to do / do not know that they should do that)
 
#### Testing Instructions
install new version with database updates on older version, in the installer you should see the messages that the database was updated. No need to go to the Kunena dashboard.
Check to see if updates are indeed done in the database.

test updates from previous version and from older version.

### DIFF
--- a/src/admin/script.php
+++ b/src/admin/script.php
@@ -16,8 +16,11 @@ defined('_JEXEC') or die();
 use Joomla\CMS\Factory;
 use Joomla\CMS\Installer\Adapter\ComponentAdapter;
 use Joomla\CMS\Installer\InstallerScript;
+use Joomla\CMS\Language\Text;
 use Joomla\Database\ParameterType;
 use Kunena\Forum\Libraries\Config\KunenaConfig;
+use Kunena\Forum\Libraries\Forum\KunenaForum;
+use Kunena\Forum\Libraries\Install\KunenaModelInstall;
 
 /**
  * Kunena package installer script.
@@ -166,6 +169,12 @@ class Com_KunenaInstallerScript extends InstallerScript
                 // remove cached Kunena Config settings
                 KunenaConfig::getInstance()->reset();
             }
+
+            try {
+                $this->upgradeDatabase();
+            } catch (\Exception $e) {
+                Factory::getApplication()->enqueueMessage($e->getMessage(), 'error');
+            }
         }
     }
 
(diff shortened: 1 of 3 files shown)


xillibit:
I have started to work in my branch related to upgrade issues : https://github.com/xillibit/Kunena-forum/tree/K7.0_add_a_function_under_tools_to_fix_socials_v2 to add a function under tools which when user choose it, it will update the socials in table #__kunena_users in case it has been failed during update process

Ruud68:
Nice! I think that moving the upgrade from the dashboard to the installer will prevent future issues where people do e.g. multiple updates without going to the dashboard in between.

xillibit:
Tested your changes by installing it on a Kunena 6.4.8, the update has right done at the end of install 👍  : 

image

The on the dashboard it shows no updates : 

image

rich20:
I tested it with Joomla 5.4.3-rc2-dev & Kunena 6.4.4 -> K6.4.9 -> K7.0.2-DEV. The database upgrade is now displayed in the installation menu and all Kunena configurations were retained.
<br />
Bildschirmfoto vom 2026-02-17 13-38-04

Ruud68:
@xillibit , posting here: maybe an idea to add some table fields in the next version with a 'sillenterror', like the language column?
That way it will be added when it is missing instead of finding out when user reported it due to not working.
Maybe go back some versions and re-add the fields for the next version. Likely we will 'catch' some issues that will then resolve themselves. Just a thought, I think that if we do not handle this 'proactively we will keep having these issues going forward with users feeling more confident to do the upgrade

xillibit:
@Ruud68 : good idea, i will change the queries by adding a 'sillenterror', 

there are already silenterror : https://github.com/Kunena/Kunena-Forum/blob/K7.0/src/admin/install/kunena.install.upgrade.xml#L579

Ruud68:
Maybe it could be as simple as changing the version numbers for the upgrade elements in kunena.install.upgrade.xml.
This file can IMO also be cleaned from a lot of entries as the 7 version will not install on versions lower then 6.4.8 and as such all entries < 6.4.8 could be removed.
History is maintained in the repository (tag)

PR #9994, opened 2026-02-16
fix for broken pagination

Ruud68:
Pull Request for Issue # https://www.kunena.org/forum/k-6-4-0-support/169075-kunena-6-4-x-paginator-issue
 
#### Summary of Changes 
 Pagination was broken due to not correctly setting tooltip class (adding double class on element)

#### Testing Instructions
test if pagination is working when tooltips are on and off

### DIFF
--- a/src/site/template/aurelia/layouts/widget/pagination/item/default.php
+++ b/src/site/template/aurelia/layouts/widget/pagination/item/default.php
@@ -22,7 +22,7 @@ if ($item->base !== null) {
     // Check if the item can be clicked.
     $limit = 'limitstart.value=' . (int) $item->base;
     echo '<li class="page-item">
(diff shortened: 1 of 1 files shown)


xillibit:
@rich20 : do-you confirm that's ok ?

rich20:
It works for me with and without tooltips. However, my test is not conclusive, as it also works for me without this fix. Tested with Firefox, Chromium and Opera.

Ruud68:
it was also tested via the forum (test result in the link.
When you lok at the code change, then you will see why it was wrong (it added class="tooltip" to the element that already had the class="..." set. It should only set the tooltip inside the already present class. That is what this change does.
Some browsers just 'drop' the second class="..."  on the element, some browsers do not do that and render the element 'disabled'. At least that is what I read from the topic itself.
I could not reproduce my self as well, although I could see when dumping the link (instead of rendering it in the browser) that it had the double class.

Issue #9986, opened 2026-01-30
The email settings are now retained after an upgrade but have no effect. The email address entered in Joomla will continue to be displayed as the sender. 


xillibit:
The email settings are now retained after an upgrade but have no effect. The email address entered in Joomla will continue to be displayed as the sender.  Do we need a new ticket for that?

_Originally posted by @rich20 in https://github.com/Kunena/Kunena-Forum/issues/9985#issuecomment-3825282718_

xillibit:
It can get too the mail from this field under Subscriptions, @rich20  : 

<img width="2335" height="306" alt="Image" src="https://github.com/user-attachments/assets/e60ea760-0b2c-4119-ae0d-067457fd11e1" />

rich20:
> It can get too the mail from this field under Subscriptions, @rich20 :

Yes, but this option is disabled as default.

Ruud68:
I think in the code the snake case names are still used: this should then be changed so the new names will be used. I only did the import, not the use.
Currently offline so cannot check

Ruud68:
I just checked and in the code (K7) all correct config settings are being used: so if it doesn't work now, then it is also not working in K6.4.8
Note that there is a bug introduced in latest Joomla regarding email and sender settings: so maybe this is the cause of it not working.
More info here: github.com/joomla/joomla-cms/issues/46643

Ruud68:
Is this 'fixed' with Joomla 6.0.3 released yesterday? Can somebody who was able to reproduce this test this?

rich20:
When I enter an email address in Kunena, I no longer receive any emails at all with Joomla 6.0.3 and Kunena 6.4.9.
With Joomla 6.0.3 and Kunena 7.0.1, I do receive the emails, but the sender address is that of the Joomla setting.

Ruud68:
On my VPS for e.g. onlinecommunityhub.nl I can only send email with from [name]@onlinecommunityhub.nl, if i try to set the sender to a different domain, then my mailserver just doesn't sent the emails: this is a security measure implemented, maybe that is the case?
Can you try to configure kunena with an email adress other then the one configured in Joomla global config but with the same domainname?

rich20:
Kunena 6.4.9
I have now set up an additional email address on my domain. The message “Undelivered Mail Returned to Sender” no longer appears, but mail is still not being delivered.

Kunena 7.0.1
[...]

PR #9996, opened 2026-02-19
Fix: email sender not set when configured

Ruud68:
Pull Request for Issue #9986 . 
 
#### Summary of Changes 
fix for getting the configured email address and fallback to global config mail address

#### Testing Instructions
in Kunena Config:
1. set email to empty > email should be send by Joomla global config email sender
2. set email to 'custom' email address > email should be send by this configured email address

### DIFF
--- a/src/libraries/kunena/src/Config/KunenaConfig.php
+++ b/src/libraries/kunena/src/Config/KunenaConfig.php
@@ -328,9 +328,7 @@ class KunenaConfig
             }
 
             if ($key === 'email') {
-                $email = $this->email;
-
-                return !empty($email) ? $email : Factory::getApplication()->get('mailfrom', '');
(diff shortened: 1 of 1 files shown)


rich20:
- [x] Success

rich20:
What is the “Sender name for mail” option for? The name does not appear anywhere in a sent email.
<br />
image111111

xillibit:
It's passer to setSender() in the second parameter : https://github.com/joomla/joomla-cms/blob/805f3f8c35cc49cc985072cc56752b24908ce4cf/libraries/src/Mail/Mail.php#L197 it's problably displayed next to the address mail

Ruud68:
fixed sender name here: #9999 

not sure why this is needed, when time permits i will research further as this might have other consequences as well.
For now this is fixed.

PR #9999, opened 2026-02-20
Fix: set sender name from configuration

Ruud68:
Pull Request for Issue #9996 . 

also for issue: https://www.kunena.org/forum/kunena-7-0-0-support/169081-email-kunena-logo-image-suddenly-disappeared-after-update
 
#### Summary of Changes 
the empty function doesn't handle the config get correct resulting in an empty = true while a string was returned by config
this resulted in the site sender name being used instead of the configured one.

#### Testing Instructions
set Sender name for mail to a name and check if that name is correctly displayed in the receid email next to the sender's email

### DIFF
--- a/src/libraries/kunena/src/Config/KunenaConfig.php
+++ b/src/libraries/kunena/src/Config/KunenaConfig.php
@@ -303,6 +303,19 @@ class KunenaConfig
     }
 
     /**
+     * Magic method to check if the key exists in the protected array.
+     * This allows empty() and isset() to work correctly.
+     *
+     * @param   string $key The key requested (e.g., 'name' for $class->name)
+     * @return  bool true if the key exists
+     * @since   7.0.2
+     */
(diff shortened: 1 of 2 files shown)


Ruud68:
Found the root cause, for the records:

We get the values from the KunenaConfig class via a magic method (__get), this allows for the direct calling of the key e.g. $config->emailSenderName, this would then use the __get function to get the value for the provided key.

The issue here is with the isset and empty methods. These do not use the __get magic method and such fail, so even while the value for emailSenderName was set, empty would still give a true (empty).

The solution is to add another magic method in KunenaConfig: __isset()

both empty and isset will first call this method and based on the result (when true), empty will then use the __get method to get the value and will then correctly state false (as it is not empty).

So this is the root cause fix for this issue

Leen29:
Thanks Ruud,
I applied your fix and it seems to work fine on
my test site.

PR #10004, opened 2026-02-25
 Commits on Feb 25, 2026      Fix to avoid issue class EasySocialPlugins is not found 

xillibit:
Pull Request for Issue # . 
 
#### Testing Instructions


rich20:
- [x] the error is now gone if EasySocial is not installed

Ruud68:
Question is also: when EasySocial is installed (J5 / J6) does the plugin then work correct? Maybe do a shout-out on the forum to get somebody who actually uses EasySocial to test? Same basically applies as well to the other plugins I think.

xillibit:
I have answered to the following topic, if someone want to test : https://www.kunena.org/forum/kunena-7-0-0-support/169080-class-kunena-forum-plugin-kunena-easysocial-extension-easysocial-not-found?start=6#234610, i will test later today on local install with easysocial

PR #10008, opened 2026-03-04
[K6.4] Fix avatar path and update username display logic in community plugin

zinshuke:
### Steps to Reproduce

* Install JomSocial and Kunena
* Enable the `plg_kunena_community` plugin
* Allow guests to comment
* Create a topic
* Post a comment as a guest
* Log in as an existing user
* Return to the topic created above
* Notice that the guest’s comment shows an incorrect avatar and displays the currently logged-in user’s name

### Summary of Changes

* Fixed the avatar link and display name for guest users

### Testing Instructions

* A guest’s comment should display the default avatar from JomSocial and the correct guest name


xillibit:
K6.4 is in End Of Life so no need to make PR against it

Issue #10012, opened 2026-03-05
Configuration report cannot be selected for copying


rich20:
**Describe the bug**
https://www.kunena.org/forum/general-questions-and-how-tos/169091-configuration-report-cannot-be-selected-for-copying
Confirmed

**To Reproduce**
Go to Kunena -> Tools -> Report Configuration 
Click on Select all and copy 

**Result**
Nothing is copied
The console displays
```
Uncaught ReferenceError: jQuery is not defined
https://www..../administrator/index.php?option=com_kunena&view=tools&layout=report:55
```

**Additional context**
Of two of my seemingly identical test pages, the bug only occurs on one of them. According to user Gindi, this error already occurs in K6.4.10.

**System information**
- Joomla version: 5.4.4-dev
- Joomla template: CASSIOPEIA
- Kunena version: 7.0.2-DEV
- Kunena template: Aurelia
- Php version: 8.3.29 & 8.4.16
- Database version: 10.11.14-MariaDB

**Desktop**
- OS: Ubuntu 24.04.3
- Browser: Chromium, Firefox

PR #10014, opened 2026-03-06
Replace broken jQuery with  working vanilla JS for copying reports to…

Ruud68:
… clipboard

Pull Request for Issue #10012 . 
 
#### Summary of Changes 
replace broken jQuery with vanilla JS and modern browsers clipboard function

#### Testing Instructions
test in Tools > Report Configuration if both buttons work and copy content (which will be displayed as selected) to the clipboard

### DIFF
--- a/src/admin/tmpl/tools/report.php
+++ b/src/admin/tmpl/tools/report.php
@@ -23,35 +23,44 @@ $wa = $this->document->getWebAssetManager();
 $wa->usePreset('choicesjs')
     ->useScript('multiselect')
     ->addInlineScript(
-        "
-		jQuery(document).ready(function ($) {
-	$('#link_sel_all').click(function(e) {
-		$('#report_final').select();
-		try {
-			var successful = document.execCommand('copy');
-			var msg = successful ? 'successful' : 'unsuccessful';
-			console.log('Copying text command was ' + msg);
-		}
-		catch (err)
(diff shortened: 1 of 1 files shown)


Leen29:
Hi Ruud,
I tested your fix too, and it seems to work.
Only on our forum when I added the report to a topic.
After posting, the editor immediately went to a new message to be cleared.
But the previously created configuration report topic did appear after refreshing.

Ruud68:
@Leen29 the fix only fixes the selection of the text and put it in the computer's clipboard, it only works on that specific view for copying the data.
What you then do with the data (e.g. past it in a topic), etc. is not changed as this has no relation with this PR.
So not sure how to troubleshoot this :(

Leen29:
Yeah, I don't think it's because of your fix either.
But I think it's a setting issue on our own forum.

Leen29:
Ruud,
I also tested your fix in our original forum
with Joomla 5.4.3 and Kunena 7.0.2, and it works fine there too.
p.s.
My previous test problem was on a test site with Joomla 6.0.3
and Kunena 7.0.2, and also with JDownloads, and it's not yet compatible with Joomla 6, so that's probably why I had my previous problem with the editor.

Ruud68:
thanks for testing and reporting back @Leen29

Issue #10017, opened 2026-03-08
Kunena 7.0.2 upgrade results in post attachments no longer working


makitso:
**Describe the bug**
With this version of kunena you can successfully add a file to your post.  However, after that the system can't find it to display.
Older posts with attachments fail to display those attachments.

**To Reproduce**

**Expected behavior**

**Actual result**
A clear and concise description of what you see. Like an error message.

**Screenshots**

**System information (please complete the following information)**

Joomla version: 6.03
Kunena version: 7.0.2
Php version: 8,3,30
Database version: Mariadb 10.11.14

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

makitso:
Reverting Router.php to 7.0.1 resolved the problem.

c-schmitz:
I can confirm the issue.,

Ruud68:
can you try the fix in PR #10019

makitso:
I am sorry but that fix did not solve my problem.

Ruud68:
with the steps to reproduce above it is 'guessing' for me. Can you tell me what you do and where you get the error?

makitso:
If I open an existing topic/post that the author had attached a file to, the display has an unknown icon.  If I select the icon I get a file not found error.
If I create a new post and attach a file to be imbedded in the post, everything works correctly and I see the attached jpeg.  If I close out and then look at my post again I can't see the attached file.

makitso:
Turn off Protect Attachments, clear browser cache and then the attachment shows.

<img width="508" height="218" alt="Image" src="https://github.com/user-attachments/assets/7158d3b3-61f8-4874-9987-c1f73777f074" />

Ruud68:
~I can't reproduce: this is what I do, please correct me if these are not the correct steps:~

Okay, so the issue is with protected attachments: these are being build using the router.

to reproduce
1. in kunena config set Protect Attachments to YES
2.  login to front-end
3. click [New Topic]
4. select category, subject, message
5. click [Add File] and select an image > click [submit]

the image is added below the message,

6. logout and login with other user
7. goto the post just created
8. the image thumb and icon is displayed incorrect
9. clicking the image gives 404

PR: #10020

rich20:
> 8. the image thumb and icon is displayed incorrect

can't confirm this error, the image thumb is correct, no error when click on it. But I have another error. 

1. in kunena config set Protect Attachments to YES
2. login to front-end
[...]

Issue #10013, opened 2026-03-05
Attachments are not displayed if Protect Attachments = Yes


rich20:
**To Reproduce**
Upload attachments with and without protected attachments

**Actual result**
If you switch the configuration to Protect Attachments = Yes, all attachments are hidden.

**System information**
- Joomla version: 6.0.4-dev & 5.4.4-dev
- Joomla template: CASSIOPEIA
- Kunena version: 7.0.2-DEV
- Kunena template: Aurelia
- Php version: 8.3.29 & 8.4.16
- Database version: 10.11.14-MariaDB

**Desktop**
- OS: Ubuntu 24.04.3
- Browser: Firefox, Chromium

makitso:
Setting Attachments = no results in no attachment in the body of the post.   Attachments=yes shows attachment in body of post but 404 when trying to open.

[confidential][b]Joomla! version:[/b] 6.0.3 [b]Platform:[/b] Apache (www.webtrees.net) [b]PHP version:[/b] 8.3.30 | [u]mbstring:[/u] Enabled | [u]GD:[/u] bundled (2.1.0 compatible) | [u]openssl:[/u] Enabled | [u]json:[/u] Enabled | [u]fileinfo:[/u] Enabled | [b]MariaDB/MySQL version:[/b] 10.11.14-MariaDB-0+deb12u2-log (Server type: mysql) | [b]Base URL:[/b]https://www.webtrees.net/[/confidential][quote][b]Database collation check:[/b] [color=green]The collation of your table fields are correct[/color]
		[/quote][quote][b]Joomla! SEF:[/b] Enabled | [b]Joomla! SEF rewrite:[/b] Disabled | [b]FTP layer:[/b] Disabled |

Ruud68:
test #10020 as this is related (I think)

makitso:
With Protect Attachment = YES, I see:

<img width="339" height="145" alt="Image" src="https://github.com/user-attachments/assets/18e5f5d2-5eb5-473c-b195-8babdde18a64" />

If I select the above image I get:

<img width="446" height="129" alt="Image" src="https://github.com/user-attachments/assets/b943aba6-581c-4820-b929-1ba7813b72db" />

IMO, not a very informative message.

Ruud68:
@makitso the second image "The requested content..." is actually a place holder as clicking the image gives a 404. You can check that by opening the browser console, clearing it and clicking the image. You get the popup as displayed but you see that the url requested gives a 404.

[...]

PR #10019, opened 2026-03-09
Fix: attachment throws 404 when added to topic but not inserted into …

Ruud68:
…message

Pull Request for Issue # descibed here . 

1. PHP version 8.5 (could also be issue for 8.4), Joomla 6.0.3
2. in Joomla Global Configuration, set Error Reporting to Maximum
3. login in on front-end
4. in Kunena Forum, click [New Topic]  in menu navbar
5. fill in / select:
 6. category
 7. subject
 8. message
9. click [Add Files] and select a file, e.g. an image
10. do NOT insert this in message and do not set it as private! but directly click [submit]

you will directly get an error 404 because the URI in the browser redirected to after submit has deprecated messages in it and as such cannot be parsed correctly
```
/forum/<br /><b>Deprecated</b>:%20 Using null as an array offset is deprecated, use an empty string instead in <b>/var/www/vhosts/developmenthub.nl/och.developmenthub.nl/j4/libraries/kunena/src/Route/KunenaRoute.php</b> on line <b>324</b><br /><br /><b>Deprecated</b>:%20 Using null as an array offset is deprecated, use an empty string instead in <b>/var/www/vhosts/developmenthub.nl/och.developmenthub.nl/j4/libraries/kunena/src/Route/KunenaRoute.php</b> on line <b>335</b><br />/j4/forum
```

#### Summary of Changes 
in the KunenaRoute.php we get the view needed to build the URI, but because the view is missing the get defaults to null. we do not check this and try to use null as index key which will give the deprecated messages.

#### Testing Instructions
See case above.

Initially Ithought this was the issue posted by @makitso and @c-schmitz but this turned out to be a different not yet raised issue.

pinging also @xillibit and @rich20

### DIFF
--- a/src/libraries/kunena/src/Route/KunenaRoute.php
+++ b/src/libraries/kunena/src/Route/KunenaRoute.php
@@ -289,7 +289,7 @@ abstract class KunenaRoute
         static $candidates = [];
         KunenaProfiler::getInstance() ? KunenaProfiler::instance()->start('function ' . __CLASS__ . '::' . __FUNCTION__ . '()') : null;
 
-        $view   = $uri->getVar('view');
+        $view   = $uri->getVar('view', '');
         $catid  = (int) $uri->getVar('catid');
         $Itemid = (int) $uri->getVar('Itemid');
         $key    = $view . $catid;
--- a/src/site/src/Service/Router.php
+++ b/src/site/src/Service/Router.php
@@ -374,7 +374,9 @@ class Router extends RouterView
                     $vars    = $variables + $vars;
                     continue;
                 } else {
-                    break;
+                    if ($segment !== 'attachment') {
+                        break;
+                    }
                 }
             }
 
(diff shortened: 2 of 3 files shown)


Ruud68:
Okay, so this is a combined PR that addresses several issues reported (#9969 , #10013 , #10017, #10020)
It would greatly help if we can test this as a solution against all related issues because it is getting more difficult to track what has been tested and how, and if things are fixed as while testing also new issues are reported in the issue...

pinging @xillibit @rich20 @makitso @c-schmitz

rich20:
If @c-schmitz, @makitso, and @xillibit no longer have any problems, then everything is fine. Even images that I added over two years ago are displayed correctly for me.

xillibit:
I'am on Php 8.4.18 on Joomla! 6.1-DEV with protected attachments set to on without this PR, the image in attachments can't be displayed. By applying this PR the image in attachments is right displayed.

> you will directly get an error 404 because the URI in the browser redirected to after submit has deprecated messages in it and as such cannot be parsed correctly
> 
> ```
> /forum/<br /><b>Deprecated</b>:%20 Using null as an array offset is deprecated, use an empty string instead in <b>/var/www/vhosts/developmenthub.nl/och.developmenthub.nl/j4/libraries/kunena/src/Route/KunenaRoute.php</b> on line <b>324</b><br /><br /><b>Deprecated</b>:%20 Using null as an array offset is deprecated, use an empty string instead in <b>/var/www/vhosts/developmenthub.nl/och.developmenthub.nl/j4/libraries/kunena/src/Route/KunenaRoute.php</b> on line <b>335</b><br />/j4/forum
> ```

I'am not able to reproduce this one with error set to max and debug mode enabled in both Joomla! and Kunena even with protected attachments set to on

PR #10020, opened 2026-03-10
Fix for #10017

Ruud68:
Pull Request for Issue #10017  . 
 
#### Summary of Changes 
When adding an attachment AND kunena is configured to 'protect attachments = yes', the the attachments gets a sef URL, the build for this url failed due to earlier change. This PR fixes that.

#### Testing Instructions

Do not only test the issue as described in https://github.com/Kunena/Kunena-Forum/issues/10017#issuecomment-4031177476

But test all scenario's where URLs are build for kunena, so topics, links, user settings, attachment, announcements, etc.

### DIFF
--- a/src/site/src/Service/Router.php
+++ b/src/site/src/Service/Router.php
@@ -374,7 +374,9 @@ class Router extends RouterView
                     $vars    = $variables + $vars;
                     continue;
                 } else {
-                    break;
+                    if ($segment !== 'attachment') {
+                        break;
+                    }
                 }
             }
 


Leen29:
I also have a website where, when uploading an attachment after updating Kunena 6.x.x > Kunena 7.0.1 (which still works then) to Kunena 7.0.2, it stops working.

Before I can press the upload attachment button, it already says "Upload failed".
even when I apply the fix
kuena702uploadfailed

Ruud68:
@Leen29 can you tell me what the settings are related to attachments in that setup and if you have debugging off and error reporting to none?

Need to be able to reproduce in order to fix

Leen29:
Hello Ruud,

I have the setting as in the 2 image upload.
ps
upload1
upload2

I can also give you access to the website where it occurs if I update it to Kunena version 7.0.2.
However, I don't know how to safely give you the login details here.

Ruud68:
@Leen29 can you try with this PR @xillibit  did yesterday? https://github.com/Kunena/Kunena-Forum/pull/10023

Leen29:
Okay, test on 19-03-2026
With Kunena SEF on, it does not work with and without suggested adjustments
but also with SEF off, then it does work in both cases, so with and without the suggested adjustments..

So it could have something to do with Kunena SEF?

Maybe this test will help you a bit further
kunena_oorzaakgevonden
kunena_sef_aan
kunenasef_uit_werkt_upload

Ruud68:
@Leen29 can you mail me at info at onlinecommunityhub.nl your site where i can see this? I have tried all settings but i am not able to reproduce this on any of my sites...

PR #10023, opened 2026-03-18
Try to revert changes about attempt to fix error 500 when URL not correct

xillibit:
Pull Request for Issue #10017. 
 
#### Summary of Changes 
 
Revert partially of changes on this PR : https://github.com/Kunena/Kunena-Forum/commit/3647c58927205afabdb79b06c44962fc1736392e

#### Testing Instructions

### DIFF
--- a/src/site/src/Service/Router.php
+++ b/src/site/src/Service/Router.php
@@ -217,12 +217,11 @@ class Router extends RouterView
             $id = (int) $query['id'];
 
             if ($id) {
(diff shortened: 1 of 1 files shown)


Leen29:
Okay, I have now applied this entire fix to my problem website where I had to disable Kunena SEF because uploading attachments otherwise wouldn't work.

With this fix, I can enable Kunena SEF again, and uploading attachments etc. works properly again now.
Attaching as private also works.

PR #10041, opened 2026-04-24
[K7.0] Prevent in some cases when saving a category to the iconset value to be null

xillibit:
Pull Request for Issue # . 

See : https://github.com/Kunena/Kunena-Forum/pull/10037
 
#### Testing Instructions

### DIFF
--- a/src/admin/src/Controller/CategoryController.php
+++ b/src/admin/src/Controller/CategoryController.php
@@ -185,6 +185,10 @@ class CategoryController extends KunenaController
                 $ignore = array_merge($ignore, $access);
             }
 
+            if(!isset($post['iconset'])) {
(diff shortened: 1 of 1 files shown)


genr8r:
Tested on Kunena 7.0.6 / Joomla 6, PHP 8.4. The save-side default works correctly for newly saved categories. One note for anyone hitting this after upgrading: categories that already had iconset = NULL in the database before this fix still trigger the render TypeError, since this PR only prevents new nulls on save and doesn't backfill existing rows. A one-time UPDATE #__kunena_categories SET iconset = 'default' WHERE iconset IS NULL clears it. Thanks for sorting this out, and sorry for the slow confirmation.

PR #10059, opened 2026-05-04
Fix: avatars in mention should be limited in size

Ruud68:
Pull Request for Issue # . 
 
#### Summary of Changes 
When fetching an avatar, a size can be requested. The returned avatar will then be in that size.
But not all avatar providers have the option of returning resized images and as such will return a different size avatar.

This can be (and is) handed by adding width and height to the img element displaying the avatar.

When in a message reply and doing a mention however, the width and height is not set to the avatar and as such the mention popup can have very large images.

This change set the with and height of the avatar in the mention list to 36 (the image size actually requested for this avatar.
 
#### Testing Instructions

install this PR, do a mention and check if the avatars still display correct, also check in the source code via inspect if the width and height of the avatar in the mention list is set to 36 via the attributes on the img element.

### DIFF
--- a/src/site/template/aurelia/assets/js/edit.js
+++ b/src/site/template/aurelia/assets/js/edit.js
@@ -8,41 +8,45 @@
  **/
 
 jQuery(document).ready(function ($) {
-    const qreply = $('.qreply');
-    const editor = $('#editor');
+    const qreply = $(".qreply");
+    const editor = $("#editor");
 
-    if(Joomla.getOptions('com_kunena.pollcategoriesid') != undefined){
-		var pollcategoriesid = jQuery.parseJSON(Joomla.getOptions('com_kunena.pollcategoriesid'));
+    if (Joomla.getOptions("com_kunena.pollcategoriesid") != undefined) {
+        var pollcategoriesid = jQuery.parseJSON(
+            Joomla.getOptions("com_kunena.pollcategoriesid"),
+        );
     }
 
-    const pollexist = $('#poll_exist_edit');
-    const pollcatid = jQuery('#poll_catid').val();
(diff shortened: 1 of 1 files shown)


Ruud68:
@xillibit yesterday got notification that this was not working correct, but that seems to be deleted? Did you find an issue with this PR?

xillibit:
I have deleted my comment it was related to mentions, i'am not able to test it because i don't remember where to change the size of avatar

Ruud68:
that explains it :)
What I saw in the image in your deleted post was that you tried to mention in a new topic: this results in an empty mention list as the people filtered to mention are bound to the topic you are in (new topic is o people to mention).

When you upload an avatar using plg_kunena_kunena, there will be several thumbnails with different sizes stored. But not all avatar providers have that option, so the avatar is then stored in e.g. a 200x200pixels format.

In all places where the avatar is used, the avatar is fetched with the width / height, and the fetched avatar url is then displayed in HTML with additional width / height set on the . plg_kunena_kunena will return a 36px wide image when fetching the avatar for the mention. But the mention is the only place where the avatar is placed in the  element without the width and height set.

So the kunena avatar will display as 36px wide as that is what is fetched, but when using another avatar provider that returns e.g. a 200px wide avatar, the will then be displayed as a 200px wide avatar in the mention list resulting in a very big image / mention list.

This PR changes this as it adds the missing width / height parameters also to the  in the mention list.

The test is simple: when you have an avatar set and do a mention for that user, does it display correct (36px) in the mention list popup?

Before:
image
After:
image

xillibit:
I have just tried with an user which has right an avatar : 

image

The avatar size is correct by inspecting the code : 

image

PR #10069, opened 2026-05-17
[K7.0] Url of ajax query to upload avatar doesn't check the session token

xillibit:
Pull Request for Issue # . 
 
thanks to @0xBassia

#### Testing Instructions


Ruud68:
I think it would be good to check all ajax calls, better safe then sorry :)

xillibit:
Yes to be sure all need to be checked

PR #10120, opened 2026-06-18
[K7.0] Set value on category iconset on update and install #10037

xillibit:
Pull Request for Issue # . 

Related to teh pull request #10037
 
#### Testing Instructions


genr8r:
This covers the upgrade path cleanly. The backfill query plus the column default change together close both the existing-rows TypeError and the chance of reintroducing nulls on a fresh install. Confirmed against my earlier test case. Thanks for following through on this.

PR #10123, opened 2026-06-19
[K7.0] Fix double-encoded entities in topic meta description tags

genr8r:
### Summary

A topic's meta description and Open Graph / Twitter card tags double-encode HTML entities. The visible post body is fine; only the `<meta>` tags are affected, which degrades search snippets and social-share previews.

### Steps to reproduce

1. On any Joomla version, create a forum topic whose first post contains a double quote in the first ~160 characters, e.g. `Sites can show "not secure" warnings.`
2. Open the topic and View Source.
3. The meta tags render:
   ```html
   <meta name="description" content="Sites can show &amp;quot;not secure&amp;quot; warnings.">
   <meta property="og:description" content="... &amp;quot; ...">
   <meta name="twitter:description" content="... &amp;quot; ...">
   ```
   The entity is double-encoded (`&amp;quot;`) instead of `&quot;`.

### Cause

`KunenaParser::stripBBCode()` returns HTML-encoded text when its `$html` argument is `true` (the default) — the `UnHTMLEncode()` step is gated behind `if (!$html)`. The meta-building code passes that already-encoded string to `setDescription()` / `setMetaData()`, and the document layer applies `htmlspecialchars()` again when emitting the `<meta>` tag, producing the double encoding.

### Fix

Pass `$html = false` at the four meta-description call sites so `stripBBCode()` returns plain text and the document encodes it exactly once. This matches how the codebase already builds plain-text page titles and tooltips (`KunenaLayout`, Topics JSON views already pass `false`).

### Scope / risk

Two files, four one-token changes. No effect on visible post content — only `<meta>` description/OG/Twitter tags. Verified on a clean Kunena 7.0.6 install: double-encoded `&amp;quot;` count drops from 3 to 0; visible body unchanged.

### DIFF
--- a/src/site/src/Controller/Topic/Item/TopicItemDisplay.php
+++ b/src/site/src/Controller/Topic/Item/TopicItemDisplay.php
@@ -619,7 +619,7 @@ class TopicItemDisplay extends KunenaControllerDisplay
 
         $multispaces_replaced = '';
         if (!empty($this->topic->first_post_message)) {
-            $firstPostMessage = KunenaParser::stripBBCode($this->topic->first_post_message, 160);
+            $firstPostMessage = KunenaParser::stripBBCode($this->topic->first_post_message, 160, false);
             $multispaces_replaced = preg_replace('/\s+/', ' ', $firstPostMessage);
         }
 
@@ -688,7 +688,7 @@ class TopicItemDisplay extends KunenaControllerDisplay
             $multispaces_replaced_desc = preg_replace('/\s+/', ' ', $this->topic->first_post_message);
 
             if ($total > 1 && $page > 1) {
-                $small = KunenaParser::stripBBCode($multispaces_replaced_desc, 130);
+                $small = KunenaParser::stripBBCode($multispaces_replaced_desc, 130, false);
 
                 if (empty($small)) {
                     $small = $headerText;
@@ -696,7 +696,7 @@ class TopicItemDisplay extends KunenaControllerDisplay
 
                 $this->setDescription($small . " - " . Text::_('COM_KUNENA_PAGES') . " {$page}");
             } else {
-                $small = KunenaParser::stripBBCode($multispaces_replaced_desc, 160);
+                $small = KunenaParser::stripBBCode($multispaces_replaced_desc, 160, false);
 
                 if (empty($small)) {
                     $small = $headerText;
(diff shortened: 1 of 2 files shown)


genr8r:
Quick note for any reviewer passing by: the only red check here is the `labeler` workflow, which fails because a fork PR's token can't apply labels. It isn't a code failure. All five PHP test jobs (8.2 / 8.3 / 8.4 / 8.5) pass. The change is two files, four one-token edits, scoped to the `<meta>` description/OG/Twitter tags with no effect on visible post content. Happy to rebase or adjust if anything's needed.

xillibit:
> Quick note for any reviewer passing by: the only red check here is the `labeler` workflow, which fails because a fork PR's token can't apply labels.

I had tried to fix that but i didn't found what i need to change. Do-you know how it can be fixed ?

genr8r:
@xillibit Sorry for the slow reply, this one slipped past me. Happy to help.

The `permissions:` block in `.github/workflows/label.yml` is already correct, which is why nothing you changed there made a difference. GitHub overrides it.

When a workflow is triggered by `pull_request` from a **forked** repo, the `GITHUB_TOKEN` is forced to read-only no matter what the `permissions:` block requests. That is a hardcoded safety rule, not a config error. So `actions/labeler` asks to write labels, gets a read-only token, and fails. Same-repo PRs are unaffected, which is why it only ever goes red on outside contributions.

The fix is to switch the trigger:

```yaml
name: "Pull Request Labeler"
on:
- pull_request_target
```

Everything else in the file can stay as-is.

`pull_request_target` runs in the context of the base repo, so the token is writable. The usual warning about it is that it can expose write credentials to untrusted code, but that risk only materializes if the workflow checks out and executes the PR head. This one has no `actions/checkout` step at all. `actions/labeler` just reads the changed-file list over the API, so there is nothing to execute. This is the configuration the labeler action's own README recommends for fork support.

One note if you adopt it: with `pull_request_target`, the labeler config is read from the **base** branch, not the PR branch. That is usually what you want, and it means a contributor cannot alter labeling rules from inside their own PR.

xillibit:
Thanks, i have made the change : https://github.com/Kunena/Kunena-Forum/commit/2416c2182ae5be1b702743de9111e5203fa52b29

Issue #10155, opened 2026-07-06
Aurelia template: Logout widget contains invalid ID with a space and related CSS/JS dependencies should be updated


harrisonwillett:
**Describe the bug**
The Aurelia logout widget appears to include an invalid `id` value containing a space.

Affected file:

```text
src/site/template/aurelia/layouts/widget/login/logout/default.php
```

On or around line 54, the markup includes an `id` with a space in the value. HTML `id` values should not contain ASCII whitespace. This can cause problems for CSS selectors, JavaScript targeting, fragment navigation, accessibility relationships, and automated accessibility validation.

This may have been introduced because the ID is being used as a styling or scripting hook. For example, existing inherited CSS may target IDs such as:

```css
#nav-menu {
    width: 190px;
}

#nav-menu a {
    padding: 5px 10px 5px 15px;
}

#nav-menu img {
    margin-left: -5px;
}
```

If the invalid ID is connected to existing styling or JavaScript behavior, the immediate fix should not only correct the invalid ID but also validate and update any CSS or JavaScript dependencies that rely on that ID.

**To Reproduce**

1. Install or use Kunena with the Aurelia template enabled.
2. Navigate to a page where the login/logout widget is rendered.
3. Inspect the rendered logout widget markup.
4. Locate the element generated from:

```text
src/site/template/aurelia/layouts/widget/login/logout/default.php
```

5. Inspect the `id` attribute on or around the element generated from line 54.
6. Observe that the `id` contains a space.
7. Search the Aurelia template CSS and JavaScript for selectors or references that depend on the affected ID.
8. Confirm whether styling or JavaScript behavior depends on the invalid/static ID.

**Expected behavior**
The logout widget should not generate an `id` with whitespace.

The element should use a valid ID, if an ID is needed at all, and styling should be handled with a class instead of an ID.

Example pattern:

```html
<div id="kunena-logout-widget" class="kunena-logout-widget">
    ...
</div>
```

The related CSS should target the class:

```css
.kunena-logout-widget {
    /* styles */
}

.kunena-logout-widget a {
    /* link styles */
}

.kunena-logout-widget img {
    /* image styles */
}
```

Instead of relying on an invalid or fragile ID selector.

If JavaScript requires a selector, it should either use a class/data attribute for reusable behavior or a valid unique ID when targeting a specific instance.

Example:

```html
<div
    id="kunena-logout-widget-abc123"
    class="kunena-logout-widget"
    data-kunena-widget="logout">
[...]

PR #10156, opened 2026-07-06
[K7.0] Aurelia template: Logout widget contains invalid ID with a space #10155

xillibit:
and related CSS/JS dependencies should be updated 

Pull Request for Issue #10155. 
 
#### Testing Instructions


harrisonwillett (approved):
I do not see `userdropdownlogout` used in the CSS or JS this seems like a safe solution.

Issue #9751, opened 2024-10-07
The indexing will not be finalised


rich20:
**Describe the bug**
https://www.kunena.org/forum/k-6-3-0-support/168598-kunena-and-advanced-search-indexation-issue
Confirmed

**To Reproduce**
1. enable the plugin Smart Search - Kunena Forum
2. go to Components -> Smart Search -> Index
3. klick on top the button Index 

**Actual result**
The search begins, but it never ends, so you have to cancel the process yourself after an estimated time. Otherwise the window remains open with this message:
"The index tables are being optimised for the best possible performance. Do not close this window."

**Additional context**
The indexing itself works, but will not completed.

**System information**
- Joomla version: ‎5.1.5-dev
- Joomla template: CASSIOPEIA
- Kunena version: 6.3.6-DEV
- Kunena template: Aurelia
- Php version: 8.2.20
- Database version: 10.5.26-MariaDB

**Desktop**
- OS: Ubuntu 22.04.1 LTS
- Browser: Chromium, Firefox and Opera

rich20:
Should this ticket remain open? This bug has already been fixed for K7.0.0. https://github.com/Kunena/Kunena-Forum/pull/9947

xillibit:
No i will close it

Issue #9663, opened 2024-03-19
Kunena Finder plugin messes smart search links for Kunena topics


xillibit:
See : https://www.kunena.org/forum/k-6-2-0-support/168308-kunena-6-2-messes-links-in-smart-search

Related to this issue : https://github.com/Kunena/Kunena-Forum/issues/6025

Issue #6025, opened 2018-07-06
Wrong url output with smart search


rich20:
**Describe the bug**
https://www.kunena.org/forum/general-questions-and-how-tos/153024-smart-search-produce-wrong-url-error-404-component-was´t-found

**To Reproduce**
Use the smart search for Kunena
The index show links such as
index.php?option=com_kunena&view=topic&catid=8&id=7&mesid=33

But the real url of this message contains the menu item id
index.php?option=com_kunena&view=topic&catid=8&id=7&Itemid=8177#33

**Actual result**
If you use the links from Smart Search you get a Error 404
The new SEO option "Redirect the page when the itemid is missing" does not help in this case.

**Additional context**
If you disable Direct Component Access, you get no error but it goes always to the beginning of the topic.

**System information (please complete the following information)**

Joomla version: 3.8.11-dev
Kunena version: 5.1.2-DEV
Php version: 7.0.25 and 7.2.1
Database version:

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

PR #10162, opened 2026-07-10
[K6.4] Fix Smart Search finder plugin: abort on stray output, endless progress loop, 404 result links

ganzziani:
Fixes three independent bugs in the Smart Search finder plugin (`plg_finder_kunena`), found while debugging why Smart Search could not index my forum. All three fixes have been running on my production site (Joomla 5.4.7, Kunena 6.4.12, PHP 8.x) where indexing now completes and result links work.

### 1. Indexing aborts with "A finder plugin has had an error"
com_finder treats **any** output produced during a batch as a plugin failure (`IndexerController::batch()` checks the output buffer). The BBCode parser echoes `No eBay APP ID and/or Cert ID defined in Kunena configuration` while `KunenaParser::stripBBCode()` processes a post containing an `[ebay]` tag without configured API keys, so one old post with a dead eBay embed aborts the whole run with no useful log entry. The plugin now buffers output around `getItems()`/`index()` and logs it instead. (The deeper fix - the parser never echoing - would live in the BBCode library; this makes the indexer robust regardless.)

### 2. Progress bar never reaches 100% - indexer loops forever (root cause of #9751)
`getContentCount()` returned `MAX(id)`, but the com_finder admin JS loops until the accumulated per-batch offsets (actually indexed items) reach `totalItems`. On any forum with deleted messages (id gaps), the inflated total is unreachable: the bar freezes (~85% on my site) and the browser fires batch requests indefinitely (~5/s), even though the server finished. Now returns `COUNT(*)` and keeps the id-based query cursor in a separate `last_id` state field while counting offsets per indexed item, like the core Joomla adapters. Note: K7.0 already reworked this with `COUNT(*)` + `setLimit($limit, $offset)`; this is the equivalent fix for the K6.4 series (the id-cursor keeps the `id > n` query, which needs no `ORDER BY` and stays stable while rows are deleted).

### 3. Search result links can 404
`createIndexerResult()` baked the `Itemid` from `KunenaRoute::fixMissingItemID()` into the stored URL, and the call `$message->getUrl($message->catid, 'last', $itemid)` does not match the signature `getUrl($category = null, $xhtml = true, $itemid = 0)` (the string `'last'` lands in `$xhtml`). On my site every Kunena result 404'd because the baked-in menu item does not route the topic view; the same URL without `Itemid` works. The plugin now stores the permalink form `index.php?option=com_kunena&view=topic&catid={catid}&id={thread}&mesid={msgid}`, which Kunena resolves (page + scroll position) at request time. Related: #9663, #6025.

Bugs 1 and 3 also affect K7.0 (`src/plugins/plg_finder_kunena/src/Extension/Kunena.php` still uses `fixMissingItemID()`/`getUrl(..., 'last', ...)` and has no output capture) - happy to port this PR there if wanted.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/plugins/plg_finder_kunena/kunena.php
+++ b/src/plugins/plg_finder_kunena/kunena.php
@@ -515,9 +515,11 @@ class PlgFinderKunena extends Adapter
         // Set title context.
         $item->title = $message->subject;
 
-        // Build the necessary url, route, path and alias information.
-        $itemid = KunenaRoute::fixMissingItemID();
-        $item->url   = $message->getUrl($message->catid, 'last', $itemid);
+        // Build the necessary url, route, path and alias information. Use the
+        // message permalink (mesid) without a hardcoded Itemid: Kunena resolves
+        // the page and position at request time, while menu ids baked into
+        // stored urls can turn into 404s.
+        $item->url   = 'index.php?option=com_kunena&view=topic&catid=' . (int) $message->catid . '&id=' . (int) $message->thread . '&mesid=' . (int) $message->id;
         $item->route = $item->url;
         $item->alias = KunenaRoute::stringURLSafe($message->subject);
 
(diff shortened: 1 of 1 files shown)


xillibit:
Thanks for this PR, the branch K6.4 is only for security fixes so you should make your PR against K7.0

xillibit:
The parts which wasn't presents has been added into K7.0, i'am closing this PR

Issue #9730, opened 2024-09-02
Cancel button does not remove database entries from attachments


rich20:
**Describe the bug**
https://www.kunena.org/forum/k-6-3-0-support/168554-strange-error
Confirmed

**To Reproduce**

1. Create post
2. Upload Picture
3. Press Insert in message
4. Cancel post
5. Go to Attachment Manage and try delete this attachment

**Actual result**
Cannot save topic: Category 0 doesn't exist! 

**Additional context**
The reason is that it does not remove the database entry with the cancel button as it used to. This results in an orphaned database entry.

**System information**
- Joomla version: ‎5.1.5-dev
- Joomla template: CASSIOPEIA
- Kunena version: 6.3.5-DEV
- Kunena template: Aurelia
- Php version: 8.2.20
- Database version: 10.5.25-MariaDB

**Desktop**
- OS: Ubuntu 22.04.1 LTS
- Browser: Chromium, Firefox and Opera

Gindi50:
Joomla 5.1.4, Kunena 6.3.4 Database: 10.6.15-MariaDB

I have just tried to reproduce the error from #9730.

1. created a new topic.

2. uploaded a picture.

3. inserted into the topic with ‘Insert’.

4. cancelled the topic with ‘Cancel’.

5. deleted the inserted image in the profile under attachment management.

Current result: Error message ’

danger

The topic cannot be saved: Category 0 does not exist! ’

But in the profile in the attachment management the image is deleted.

In the database, the image only exists once in jom5_kunena_attachments.

rich20:
@Gindi50 Does it also add an orphaned attachment entry in the diagnostics?

Gindi50:
Yes, one problem was displayed with attachmentOrphaned.

Gindi50:
I have tested again with three pictures.

After selecting and before inserting the images, the images are already in profile/attachment management and in the database at jom5_kunena_attachment.

In Kunena/Tools/Diagnosis there are now 3 problems with attachmentOrphaned.

Then I inserted the three images.

Then I removed the three images in the post and everything was gone, the text and the images. The text field said ‘true’.

In the profile/attachment management there were only two pictures left, one was deleted.

In Kunena/Tools/Diagnosis there are now 2 problems with attachmentOrphaned.

In the database at jom5_kunena_attachment there are now 2 images.

Then I cancelled the post. (With the known error message)

In the profile I could delete the two pictures (with the known error message)

[...]

PR #10174, opened 2026-07-20
K7.0 store in sessions attachments before sumbit to avoid to have attachments orphaned in database #9730

xillibit:
Pull Request for Issue #9730. 
 
#### Summary of Changes 
 
Actually when the user upload attachments there are directly stored in database, with this Pull Request there are stored in user session untill the user submit his post.

#### Testing Instructions

### DIFF
--- a/src/libraries/kunena/src/Attachment/KunenaAttachment.php
+++ b/src/libraries/kunena/src/Attachment/KunenaAttachment.php
@@ -862,20 +862,20 @@ class KunenaAttachment extends KunenaDatabaseObject
 
         if (!$this->filename_real) {
             $this->filename_real = $this->filename;
-        }
-
+        } 
+        
         if (!$this->filename || $this->filename == $this->filename_real) {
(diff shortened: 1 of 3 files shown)


Ruud68:
Hi @xillibit , just looked at this and although I understand the thinking, this will lead to issues as this is not what the session is designed for. sooner or later this will lead to load / scalability and server issues. It will also not completely solve the orphan issue and likely make it worse: what if a user attaches files but then decides to close his browser / tab? The session will be removed by garbage collection at some point in time, but the file itself is still somewhere stored on disk with no reference to it.

what would be a better solution is to store the attachments in the DB with a 'pending' status. Then when the user submits, it is a atomic switch from status pending to status active (or something) else.

Then e.g. a scheduled task can delete all stored attachments with status = pending that are 1 day old.

xillibit:
I will try with your suggestion, i didn't know that storing in session can lead to server issues. I will finish first the changes on the notifications mail with task plugin.

PR #10198, opened 2026-08-05
[K7.1] Use processNotification() method instead of sendnotification()

xillibit:
Pull Request for Issue # . 
 
#### Testing Instructions


xillibit:
@Ruud68 : now it use the new method processNotification() which by default store in queue instead of sendnotification() which is now deprecated, i have tested the processNotification() it works to send the notification in real time it works in this case.

Issue #10218, opened 2026-09-01
Rendering Error in User/Profile when user email is null


fourzero0:
Kunena 7.0.8 throws a rendering error in the User/Profile layout when the KunenaUser email property is null.

Error:

Rendering Error in layout User/Profile:
Kunena\Forum\Libraries\User\KunenaUser::escape():
Argument #1 ($var) must be of type string, null given,
called in libraries/kunena/src/User/KunenaUser.php on line 1920

The affected code is:
```
case 'email':
    return '<a href="mailto:' . $this->email
        . '" data-bs-toggle="tooltip" data-placement="right" title="'
        . $this->escape($this->email) . '">'
        . KunenaIcons::email() . '</a>';
```
In my case this occurs for an old forum post whose Joomla user account no longer exists.

The message still exists in #__kunena_messages and originally referenced userid 13151.
The corresponding user existed in #__kunena_users but not in #__users.
After running Kunena user synchronization, the orphaned record was removed from #__kunena_users, but the rendering error remained.

Actual result

Rendering of User/Profile stops at the email profile icon.
Because the layout terminates before the surrounding div elements are closed, the message content column becomes nested inside the profile/sidebar column and the whole post layout is broken.

Expected result

If the user has no email address, the email profile icon should simply not be rendered and the rest of the profile/message layout should continue normally.

Possible fix

This local change fixes the issue:
```
case 'email':
    if (empty($this->email)) {
        return '';
    }

    return '<a href="mailto:' . $this->email
        . '" data-bs-toggle="tooltip" data-placement="right" title="'
        . $this->escape($this->email) . '">'
        . KunenaIcons::email() . '</a>';
```
System information

- Joomla: 6.1.3
- Kunena: 7.0.8
- PHP: 8.4.24
- Kunena template: Aurelia
- Joomla template: Cassiopeia

Pinkeltje:
Can confirm your issue and can confirm as well that your solutions works. I will create a pull request. Thank you for reporting this issue.

xillibit:
The PR has been merged

PR #10220, opened 2026-09-04
Prevent mailto link for empty email in KunenaUser

Pinkeltje:
Add check for empty email before returning mailto link.
#10218


xillibit:
FYI : the branch K6.4 is only for securities fixes

Issue #10227, opened 2026-09-11
Admin sidebar menu items fall back to English despite correct nl-NL translations (com_kunena.sys language cache collision)


Pinkeltje:
## Summary

With a correctly installed and valid nl-NL (Dutch) language pack for Kunena, the Kunena admin sidebar menu (Dashboard, Categories, Users, etc., rendered via Joomla's `mod_menu`) intermittently or persistently displays in English, even though:

- the `nl-NL.com_kunena.sys.ini` file is present, syntactically valid, and contains correct translations for every key involved (verified line-by-line),
- the backend language is confirmed set to `nl-NL` (`JFactory::getLanguage()->getTag()` returns `nl-NL` throughout),
- other backend text (Joomla core, other 3rd-party extensions like AcyMailing) is unaffected and displays correctly in Dutch,
- clearing all Joomla caches, disabling Kunena's own "fallback to English" setting, and ruling out OPcache/browser caching all have no effect.

**Environment:**
- Kunena 7.0.9
- Joomla 6.1.3
- Tested on both a local XAMPP-Lite (Windows) install and a LAMP-based live hosting environment

## Root cause (confirmed via debug logging)

Kunena's `Dispatcher::loadLanguage()` unconditionally loads `com_kunena.sys` very early on every admin request:

```php
protected function loadLanguage(): void
{
    KunenaFactory::loadLanguage('com_kunena', 'admin');
    KunenaFactory::loadLanguage('com_kunena.views', 'admin');
    KunenaFactory::loadLanguage('com_kunena.libraries', 'admin');
    KunenaFactory::loadLanguage('com_kunena.sys', 'admin');   // <-- this line
    KunenaFactory::loadLanguage('com_kunena.install', 'admin');
    KunenaFactory::loadLanguage('com_kunena.models', 'admin');
    KunenaFactory::loadLanguage('com_kunena.controllers', 'admin');
    KunenaFactory::loadLanguage('com_plugins', 'admin');
    KunenaFactory::loadLanguage('com_kunena', 'site');
}
```

Later in the same request, Joomla core's `mod_menu` (`administrator/modules/mod_menu/src/Menu/CssMenu.php::preprocess()`) also loads `com_kunena.sys`, once per sidebar item, to translate the item title:

```php
if (!empty($item->element)) {
    $language->load($item->element . '.sys', JPATH_ADMINISTRATOR) ||
[...]

PR #10229, opened 2026-09-11
[K7.0] Admin sidebar menu items fallback to Eng despite nl-NL by default 

xillibit:
Pull Request for Issue #10227. 
 
#### Testing Instructions

### DIFF
--- a/src/admin/src/Dispatcher/Dispatcher.php
+++ b/src/admin/src/Dispatcher/Dispatcher.php
@@ -42,8 +42,7 @@ class Dispatcher extends ComponentDispatcher
     {
         KunenaFactory::loadLanguage('com_kunena', 'admin');
         KunenaFactory::loadLanguage('com_kunena.views', 'admin');
-        KunenaFactory::loadLanguage('com_kunena.libraries', 'admin');
-        KunenaFactory::loadLanguage('com_kunena.sys', 'admin');
+        KunenaFactory::loadLanguage('com_kunena.libraries', 'admin');        
         KunenaFactory::loadLanguage('com_kunena.install', 'admin');
         KunenaFactory::loadLanguage('com_kunena.models', 'admin');
         KunenaFactory::loadLanguage('com_kunena.controllers', 'admin');
--- a/src/admin/src/View/Attachments/HtmlView.php
+++ b/src/admin/src/View/Attachments/HtmlView.php
@@ -21,6 +21,7 @@ use Joomla\CMS\Language\Text;
 use Joomla\CMS\MVC\View\GenericDataException;
 use Joomla\CMS\MVC\View\HtmlView as BaseHtmlView;
 use Joomla\CMS\Toolbar\ToolbarHelper;
+use Kunena\Forum\Libraries\Factory\KunenaFactory;
 
 /**
  * Attachments view for Kunena backend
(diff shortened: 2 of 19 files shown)


xillibit:
I let you @Pinkeltje test it

Pinkeltje:
Wow! Lot of changes. Will test tomorrow.

Pinkeltje:
Thanks for the quick turnaround on this! I gave it a thorough test across a few sessions and environments, and unfortunately the sidebar translation issue still shows up with this patch applied — both intermittently and, depending on which admin screen loads first, consistently.

Digging a bit further, I think I found a more targeted root cause, one level down from where this PR makes its changes.

## Root cause

It's in `KunenaFactory::loadLanguage()` (`libraries/kunena/src/Factory/KunenaFactory.php`):

```php
$loaded[$file] = $lang->load($file, $lookup1, null, $english, false)
    || $lang->load($file, $lookup2, null, $english, false);
```

The last argument (`$default`) is passed as `false`. Looking at Joomla core's `Language::load()`:

```php
public function load($extension = 'joomla', $basePath = JPATH_BASE, $lang = null, $reload = false, $default = true)
{
    ...
    // Load the default language first if we're not debugging and a non-default language is requested
    if (!$this->debug && ($lang != $this->default) && $default) {
        $this->load($extension, $basePath, $this->default, false, true);
    }
    ...
    foreach ($filenames as $filename) {
        if (isset($this->paths[$extension][$filename]) && !$reload) {
            $result = $this->paths[$extension][$filename]; // already tested, skip re-parsing
        } else {
            $result = $this->loadLanguage($filename, $extension);
        }
        ...
    }
}
```

`$default = true` is what makes Joomla core load en-GB *and* the active language together, atomically, in one call, marking both as correctly "tested". Passing `$default = false` opts out of that — so whichever admin entry point happens to run first (there are several beyond the 19 views touched here: the base controller's `before()`, the quickicon plugin on the Joomla dashboard, etc.) marks the active-language file as "tested" without Joomla's own en-GB pass ever having touched the matching `administrator/language/en-GB/...` path.

Later, when `mod_menu`'s `CssMenu::preprocess()` calls `$language->load('com_kunena.sys', JPATH_ADMINISTRATOR)` with Joomla's default `$default = true`, en-GB gets parsed fresh for the first time — overwriting the strings array — and then hits the "already tested" shortcut for the active-language file, so it never re-merges the translation back in.

This is also why other extensions (including ones with their own quickicon plugins) don't show this: they don't opt out of Joomla's atomic default-then-active load with `$default = false`, so the correct order gets locked in on first touch, however early that touch happens.

## Suggested fix

Change `$default` from `false` to `true` in that call:

```php
$loaded[$file] = $lang->load($file, $lookup1, null, $english, true)
    || $lang->load($file, $lookup2, null, $english, true);
```

With this, the manual `fallbackEnglish`-gated en-GB preload just above becomes largely redundant (Joomla core now handles that atomically), but I left it in place to keep the change minimal for testing.

## Testing

Tested this single-line change (with the 19 files from this PR reverted back to their pre-PR state) on:
- nl-NL backend language (Kunena 7.0.9, Joomla 6.1.3)
- a separate, unrelated test site with de-DE as backend language

Repeated navigation through the admin sidebar, including fresh incognito sessions, consistently shows the correct translation in both cases.

Given it resolves the issue with a one-line change in a single file, would it make sense to go this route instead? Happy to open a small PR with just that change if that's easier to review than reworking this one.

2026-09-12 09-16 Schermafbeelding
2026-09-12 09-05 Schermafbeelding

xillibit:
Good to have found the root cause, you can make a PR with this single line
change.

I will close my PR

Le sam. 12 sept. 2026, 09:21, Astrid van Eerd ***@***.***> a
écrit :

PR #10236, opened 2026-09-14
Fix [6.4] "Delete selected" doesn't work on Unapproved Posts

c-schmitz:
Problem: As a moderator, open Unapproved Posts, select one or more posts, choose "Delete post", and click GO — the posts were not removed. This is essential for us to combat spam, especially when bots submit hundreds of unapproved messages.

TopicsController::delete_posts() read the selected message IDs from the POST field topics, but the moderation list (e.g. the Unapproved Posts view) renders its checkboxes as posts[id]  - the same field name used by the sibling bulk actions approve_posts(), restore_posts(), and permdel_posts().

Because no topics[...] field is ever submitted from that form, delete_posts() always resolved an empty ID list, KunenaMessageHelper::getMessages([]) returned an empty array, and the method silently fell into its "no messages selected" branch — the delete loop never ran. Selected posts stayed in place with only an easy-to-miss notice shown, while "Approve selected" on the same page worked fine, since it correctly reads posts.

Fix: Read the selected IDs from posts instead of topics in delete_posts(), consistent with the other bulk actions in this controller.

The patch can be cherry-picked to 7.0 without problems.


xillibit:
I have cherry picked in K7.0 : https://github.com/Kunena/Kunena-Forum/commit/989bea8f3339ff6ec0a5dcef12e86018bb160dd5 there is a conflict that i need to fix

The branch K6.4 is only for securities fixes

PR #10244, opened 2026-09-21
[K7.0] Fix docblock @param names that do not match the signatures

darkdi:
Pull Request for Issue # .

#### Summary of Changes

I was reading the event classes and noticed `KunenaGetAvatarEvent::onSetAvatar` documents `$value` while the argument is `$avatar`, and the same in the `Private` and `Profile` siblings, so I went through the rest of the tree for the same thing. The ones worth a look beyond the renames: `plg_system_kunena` still documents `onUserBeforeSave` with the pre Joomla 5 signature of `$user`, `$isNew` and `$data` although the method now takes a single `BeforeSaveEvent`, and `plg_quickicon_kunena` documents `$context`, which is a local variable inside `onGetIcons` rather than its argument. `getUserDefaultLanguage` is typed `User $user` in the docblock but the body calls `json_decode` on it and the caller passes `$options['user']->params`, so it is a string.

Four docblocks document arguments for methods that take none at all, `KunenaKarma::save`, `KunenaKarma::getLastTimeKarma`, `KunenaConfig::getIdeProperties` and `KunenaModelInstall::cleanMailTemplates`. Those lines are removed rather than renamed. In `KunenaRoute::getTopicUrl` and `KunenaTopic::getUri` the difference is only the capital in `$Itemid`.

I left `libraries/kunena/External/Nbbc` alone, `setLogFile` has the same problem there but it is vendored. I also left the `@return boolean` on `onUserBeforeSave` as it is, the method returns void now, but that is a different thing and I did not want to mix it in.

#### Testing Instructions

Comments only, no behaviour changes. `php -l` passes on every touched file. Every name was read against the signature right below it.


darkdi:
Pushed one more line: `KunenaKarma::save` still carried `@internal param int $userid` under the `@param` I removed, and the method takes no arguments either way. I caught it after a reviewer on another PR pointed out that I was renaming names without reading the lines around them.

